Issue #20, opened 2016-10-02
Open source alternative to FMOD


akien-mga:
Partly related to #19, but would have a different resolution so opening a new issue.

To make Barony fully libre, it would be nice to have it support an open source alternative to FMOD. If you use specific FMOD features that you don't want to part with (and additionally, if you don't want to spent too much time porting the sound engine for a game you already have in production), it could be kept alongside the alternative, and the sound engine could be selected at build time via a CMake option.

Possible alternatives would be [cAudio](https://github.com/R4stl1n/cAudio) (I discovered it recently and heard many good things about it), RtAudio, OpenAL directly, SDL_mixer, SFML Audio, PortAudio, BASS... I don't have experience with most of them, but with a bit of research I'm 100% sure that you can find a great open source library that could be used as a replacement for the proprietary FMOD.

Until an alternative is implemented, it would be nice to have the possibility to disable FMOD via a CMake argument (one would thus likely lose sound support, but at least it would be possible to build without proprietary library).

addictgamer:
Moving away from FMOD is something we've discussed -- in fact, I believe we're using OpenAL or thereabouts for future projects.

As for Barony, I don't plan on porting any of the sound code at this point, but I would be more than happy if somebody took it up to port the sound code.

SheridanR:
We are indeed using OpenAL Soft (ie the free implementation of OpenAL) for our next project, partly because FMOD seems to be under a somewhat restricted license. At this stage it would be costly for me to  dive back into Barony and rewrite the sound engine, but if anyone else is interested that is definitely the library I would recommend for that purpose.

akien-mga:
Thanks for the feedback. As a first step I'll probably make a pull requests to add an optional argument that disables FMOD support (and thus the sound engine) until a replacement is contributed. I don't think many users will want to play the game without sound, but just for the sake of it :)

Regarding helping replace FMOD with OpenAL Soft or another libre alternative, I might have a look at the work it implies, though I'm probably not the best fit for the job (never implemented a sound library myself).

akien-mga:
Looks like you not only need to register an account to download FMOD Ex, but you also need to then contact support to get your account enabled for downloading it... Don't think they'll want to enable my `jdoe_onetime` junk account, so let's start with implementing a way to disable it :D

addictgamer:
Huh, that's new. It used to be you just needed an account to download (even farther back, anybody could just come and download anything they wanted off of the website). Now you need to get your account "enabled"?

[...]

PR #22, opened 2016-10-11
Abstract away the sound engine

pwhelan:
Here is the initial work to remove the dependency on FMOD. Hopefully it can serve at least as a starting point for #20.

It worked for me at least during a very simple test run.

I have not done any of the due diligence to make it cleaner. As well, there is a slight possibility some initialization code that is important outside of sound has now been tucked away inside of an '#ifdef'.


akien-mga:
From a quick test on Linux it seems to work fine, the engine compiles and the game and editor run fine (of course without sound ;)).

I did not test compile with FMOD though, only without it.

addictgamer:
Awesome, I'll begin reviewing this as time permits.

src/acthudweapon.cpp
```diff
@@ -108,8 +108,11 @@ void actHudArm(Entity *my) {
 	my->pitch = -17*PI/32;
 }
 
+#ifdef HAVE_FMD
```
addictgamer: Should be:
#ifdef HAVE_FMOD

addictgamer:
Took a look over the changes, noticed a few issues I pointed out in the comments.


On to compiling: It does not compile without FMOD for me:
game.cpp:1733:19: error: ‘introductionmusic’ was not declared in this scope
         playmusic(introductionmusic, TRUE, FALSE, FALSE);
                   ^~~~~~~~~~~~~~~~~
game.cpp:1733:56: error: ‘playmusic’ was not declared in this scope
         playmusic(introductionmusic, TRUE, FALSE, FALSE);
                                                        ^
game.cpp:1738:19: error: ‘intromusic’ was not declared in this scope
         playmusic(intromusic, TRUE, FALSE, FALSE);
                   ^~~~~~~~~~
game.cpp:1738:49: error: ‘playmusic’ was not declared in this scope
         playmusic(intromusic, TRUE, FALSE, FALSE);

Those lines should probably have been thrown into an #ifdef as well.

As for trying to compile with FMOD, once I fixed the issues in CMakeLists.txt and acthudweapon.cpp, it compiled and seemed to run just fine.
For convenience, I pushed the changes to the pull request.


Also, for future reference, I'd prefer defines be passed to the game via Config.hpp.in rather than add_definitions().

akien-mga:
@addictgamer Off-topic, but do you happen to have push rights on @pwhelan's repo, or is there a new GitHub trick that allows pushing commits in PRs on your repos?

addictgamer:
The latter: https://help.github.com/articles/committing-changes-to-a-pull-request-branch-created-from-a-fork/

akien-mga:
Thanks, that's cool to know :) `</off_topic>`

addictgamer:
Indeed it is, just found out about it earlier today :)

pwhelan:
> Also, for future reference, I'd prefer defines be passed to the game via Config.hpp.in rather than add_definitions().

@addictgamer done!

From what I glanced one of the playSoundX functions sends sounds through the network, is that right? If so, the no sound option can be implemented send out those packets for those fortunate enough to have sound. I just wanted to ask first.

addictgamer:
Those functions send the sound index. position, and volume. Those packets should be sent regardless of whether or not the host has sound enabled.

So, yes, good question, go for it.

addictgamer:
I'm having some issues with the sound disabled version:
Without Steamworks:
 * Get into the logo and pirate disclaimer just fine. The screen just fades to black after the pirate disclaimer.
With Steamworks:
 * Everything appears to be working fine, except when trying to end the game. The screen fades out to black and sits there.

In general:
 * Bows don't work anymore. It looks like the bow times itself to the sound effect. No sound = no firing bow.

addictgamer:
Alright, fixed the issues with the game hanging on fading to black. b81f5d2f00ebb28177969f08993f28a1e199170c
It looks like some of the variables that controlled the screen fading out were inside of the #ifdef, which meant no sound = the game not being able to progress.

Additionally, I threw in an FMOD_ENABLED environment variable which can be used to explicitly disable FMOD even if you have it installed on your system (otherwise, FMOD will be used if CMake can find it). 2d293938ad3a4cabe2588e0eb8b96a144afab0bd

As for the bow, that looks like it would be...interesting...to deal with.

addictgamer:
Unless any further complications arise, I think I will merge in this pull request after the playSound*() functions have been updated, with the disclaimer that the no-sound-version is not officially supported and we are not liable for anything that might break as a result :)

pwhelan:
@addictgamer are the values of playSoundX ever used?

addictgamer:
Yes, the FMOD_CHANNEL* returned is indeed used.

Good example: Bows are broken when sound is disabled because the bow uses the current status of bowDrawingSound to go through the firing sequence.

pwhelan:
I've just pushed netcode for the null sound implementation. For some reason neither right-click or left-click work inside the dungeon.

addictgamer:
Go into Settings -> Keyboard tab and make sure "Attack" and "Use" are bound.

pwhelan:
> The bow is the only part of the game I can think of that uses the state of a playing sound to carry out game logic. Hopefully there's not more...

I spotted that now as I was trying to figure out my input problem. Do you have a preferred way of handling timed events outside of this mechanism?

pwhelan:
Ok... This is a *really* bizarre bug. Somehow those key bindings do not appear in my local build.

no-use-attack

It works fine under STEAM. I also made an attempt at merging this branch with the gamepad branch (which worked, btw). Might running it have caused some kind of problem?

pwhelan:
STEAM also does not show these entries in Settings -> Keyboard.

addictgamer:
The Steam 2.0.3 Beta branch should have those entries in Settings -> Keyboard. (v2.0.2 does not)
Did you copy over lang/en.txt? It's possible it's still rendering those options in your local build...as empty strings.

pwhelan:
I basically have a script and a symlink in my build directory so I can use the STEAM assets with my build. Opting into the public beta fixed both bugs with one fell swoop (so that bit was all on me).

The basic idea with this patch is to get support for compiling without FMOD to work with the least amount of changes would an ok compromise for bows be to use SDL_GetTicks to set/unset the booleans (using a hard coded time)?

addictgamer:
>The basic idea with this patch is to get support for compiling without FMOD to work with the least amount of changes would an ok compromise for bows be to use SDL_GetTicks to set/unset the booleans (using a hard coded time)?

I would be fine with that.

pwhelan:
I just took a stab at fixing the bow. I tried it with the arcanist and it works for me, as far as I can tell.

akien-mga:
FWIW, I think an effort to further abstract the sound engine (by decoupling game logic from the sound management, and just let the logic fire sound events that would be handled by a SoundManager or similar) would make this all easier to maintain and extend. But of course, it's quite some work for no directly visible gain :)

pwhelan:
@addictgamer which character uses the bow? I think the arcanist is just using a cross bow (which probably does not trigger the bow code).

pwhelan:
@akien-mga I may take a stab at that later when I try reimplementing it with SDL_mixer.

addictgamer:
@pwhelan 
>@addictgamer which character uses the bow? I think the arcanist is just using a cross bow (which probably does not trigger the bow code).

The warrior and rogue.

>@akien-mga I may take a stab at that later when I try reimplementing it with SDL_mixer.

Actually, we were using SDL mixer up until about 2 years ago, and then we decided to move away from it. I would very much like to not go back. OpenAL or something along those lines would be preferred to SDL mixer.

addictgamer:
> FWIW, I think an effort to further abstract the sound engine (by decoupling game logic from the sound management, and just let the logic fire sound events that would be handled by a SoundManager or similar) would make this all easier to maintain and extend. But of course, it's quite some work for no directly visible gain :)

I would very much like a layer of abstraction between the sound implementation and the game code. Unfortunately, for the reason you mentioned, it's not on my personal TO-DO. Anybody who wants to take it on has my blessing.

pwhelan:
@addictgamer any good starting reference for OpenAL?

addictgamer:
> @addictgamer any good starting reference for OpenAL?

I'm afraid the extent of my OpenAL knowledge is limited to:
 * We're using OpenAL Soft ([repository](https://github.com/kcat/openal-soft) and [website](http://kcat.strangesoft.net/openal.html) -- the official OpenAL implementation is defunct as far as I'm aware) from here on out
 * and all I need to do to use the sound engine is `int handle = entity.playSound(...);`
(The sound implementation is nicely abstracted and not jammed into the guts of the game :) )

I'm sorry I can't be of any more help. Issue #20 listed some suggestions to other sound libraries you could take a look at. I don't particularly care which library is used as long as it's compatible with all of our platforms (OSX, Windows, Linux) and proves the features we need (namely 3D sound).
For what it's worth, @SheridanR is the one who's dealt with OpenAL Soft, and he favorably recommended it in the aforementioned issue.

addictgamer:
@pwhelan Is there anything else you wanted to do before I accept the pull request? I'm ready on my end.

akien-mga:
Are the 36 commits of this PR all meant to be there actually? The commits by @lakp sound like they might have been mistakenly pushed to this branch.

pwhelan:
> Is there anything else you wanted to do before I accept the pull request? I'm ready on my end.

If bows are working then I'm good too.

pwhelan:
> The commits by @lakp sound like they might have been mistakenly pushed to this branch.

That looks like a merge with master.

addictgamer:
> Are the 36 commits of this PR all meant to be there actually? The commits by @lakp sound like they might have been mistakenly pushed to this branch.

Oops. I think I know what happened there.

> If bows are working then I'm good too.

They worked in the quick test I ran.

PR #23, opened 2016-10-12
Auto appraise and quick drop

lakp:
**Auto appraise:**
Option added to misc tab in settings. (default: disabled)
If enabled, the player will automatically start appraising items if he/she has unidentified items and is not appraising anything.

**Quick drop:**
Similar to shift+right click to appraise, shift+left click drops items.


lakp:
I got tired of opening the inventory to appraise things.

addictgamer:
I'll take a more detailed look over it when I get a chance, but I did notice there will be a lang/en.txt conflict over 1961 with the gamepad branch's additions.
Can you change it to something else? I'm not sure what would be good, but perhaps 1997?
(The big language file/array should probably be split into some sort of thematic collection eventually, but until then...)

addictgamer:
Hmm, here's something interesting I noticed: Gems (especially worthless glass) can essentially get you stuck in an appraisal loop on just that item. Maybe it should try another item before re-attempting the same item.
Thoughts?

lakp:
Yeah I could make it work on the easiest items first.

addictgamer:
Sounds good to me.

lakp:
Discovered you can quick drop spells.

addictgamer:
Ah, good catch. I did not test for that.

lakp:
There's a bug where having a rock in your inventory will cause an infinite appraisal loop.

lakp:
Should be fine now.

addictgamer:
Is a map a good idea? Wouldn't keeping a pointer to the item with the shortest appraise time suffice?

It's not a major issue or anything, seeing as the map is not created every frame, but...

lakp:
Yeah that sound way better.

addictgamer:
Alright. Excepting any issues that might (hopefully not!) arise, I'll give the code a whirl after that's done and merge in the pull request.

lakp:
Done.

addictgamer:
Alright, I'll give it a shot tomorrow (or later today, at this point!), after my mental facilities have had some rest. It's well past midnight over here :)

addictgamer:
> I should probably throw together a document at some point.

Done. Funnily enough, the changes in e77021a are now representative of the project's correct coding style.

PR #50, opened 2017-01-25
Added OPENAL and PANDORA support, Drawing and Logic optimizations

ptitSeb:
There are a lot of things here, sorry for the huge and mixed pack.

There is no Sound Abstraction, so OPENAL / FMOD is not as clean as it could be.
This is the version I used to make Barony: Cursed Edition work on the OpenPandora with correct speed.


akien-mga:
That's great, but you should really have split that in several commits *and* several PRs. Maybe the devs will want to merge only part of your changes and not them all as a huge commit..

akien-mga:
Also you seem to have changed file permissions to make source files executable: `src/collision.cpp 100644 → 100755`

ptitSeb:
permissions changes comes from my Linux/Windows mix. I'll try to removes thoses changes

The splitting... well, it's a lot of work, so I try the whole pack first. If it's not accepted, I'll see how to do the split (I made this to port to the Pandora, that's why it's an all-in-one PR).

ptitSeb:
Ok, fixed the permission stuffs.

addictgamer:
Several commits and PRs would have been nice, but regardless, I hope to peruse this pull request this weekend and get back to you.

CMakeLists.txt
```diff
@@ -16,6 +18,20 @@ endif()
...
+option(OPENAL_ENABLED "Use the OpenAL library in the sound engine" OFF)
+set(TREMOR 0)
+option(TREMOR_ENABLED "Use Tremor instead of libvorbis for OpenAL" OFF)
+
+if(FMOD_ENABLED AND OPENAL_ENABLED)
+  set(FMOD_ENABLED OFF)
+endif()
+
+set(PANDORA 0)
+option(PANDORA_ENABLED "Build forthe OpenPandora" OFF)
```
addictgamer: `option(PANDORA_ENABLED "Build forthe OpenPandora" OFF)`

Just a minor typo there. Should be:

`option(PANDORA_ENABLED "Build for the OpenPandora" OFF)`

(Will not impede acceptance)

ptitSeb: Ah yes indeed!

addictgamer:
Looked over it and did some preliminary testing. Looks like almost everything is working great -- why did you change `double` to `DOUBLE`?

ptitSeb:
I used DOUBLE to be abble to use float on arm platform. double or float doesn't have much impact on x86 world, but it make a huge difference when using arm cpu.

addictgamer:
I see. `DOUBLE` is causing compilation issues in Windows. The `#define DOUBLE double` should only be triggered outside of Windows environments, as it seems the Windows API already includes a `typedef double DOUBLE`.

addictgamer:
Also, I'm not sure what's going on with this whole "new"/"old" loop thing in collision.cpp?
It's rather messy -- any particular reason why there are two loops, and all that `#ifdef` going on?

SheridanR:
The NEWLOOP define is not necessary -- if we detect problems later we can blame the new code and roll back or fix it, but right now it looks solid.

Everything else looks good as well. I appreciate that you went through and cleaned up a lot of garbage in the collision and rendering code. We did not know what the hell we were doing with OpenGL when we wrote it.

Also, huge props in general for the added OpenAL support, that's wicked awesome.

We do need to test this a little further -- in particular, trying out OpenAL (so far we have only tested that it still compiles with FMOD). But everything else seems to work great on all three platforms. So, if the DOUBLE issue is fixed as Ciprian suggested and NEWLOOP is removed (keep your new code, that's great, but remove the old stuff) then we will be totally happy to accept this pull request.

akien-mga:
Maybe `real_t` would be a better type to be able to switch between `float` and `double` based on a compiler define? A `DOUBLE` type that happens to be `float` sounds a bit misleading.

ptitSeb:
Yeah, I'll fix the DOUBLE issue (putting a typedef instead of a #define) (and yeah, I'll see to change it to real_t, why not).
I'll also remove the #define NEWLOOP no problem.

ptitSeb:
Ok, it's done.

addictgamer:
Looking good. Barring any issues, I'll merge this in the coming days after further testing.

Thanks again.

addictgamer:
The OSX situation is very weird. Looks like Apple distributes OpenAL, but the ALUT situation is very confused.
Also, error on compile:
```
fatal error: 
      'AL/al.h' file not found
```
It looks like OSX just has al.h directly, not under the AL folder.

CMakeLists.txt
```diff
@@ -127,6 +150,26 @@ else()
...
+  else()
+    find_package(OGGVORBIS)
+  endif()
+  if (OPENAL_FOUND AND ALUT_FOUND AND ((TREMOR_ENABLED AND TREMOR_FOUND) OR ((NOT TREMOR_ENABLED) AND OGGVORBIS_FOUND)))
+    INCLUDE_DIRECTORIES(${OPENAL_INCLUDE_DIR} ${ALUT_INCLUDE_DIR})
+    if(TREMOR_ENABLED)
+      set(TREMOR 1)
+      INCLUDE_DIRECTORIES(${TREMOR_INCLUDE_DIR})
+    else()
+      INCLUDE_DIRECTORIES(${OGGVORBIS_INCLUDE_DIR})
```
addictgamer: This doesn't work for me under OS X. It seems that `OGGVORBIS` is never defined. Rather, two separate variables are:
`VORBIS_INCLUDE_DIR` and `OGG_INCLUDE_DIR`
Including those two directories fixes the issue for me (or, rather, including only one is required since they're both located in the same directory).
If I had to take a guess, this works just fine on my Linux machine since their include directory is already included by practically every other library, thus hiding this issue.

addictgamer:
By the way, I did get the OpenAL code working in Linux without a hitch. I did notice one oddity, though: equivalent sound settings are quieter in OpenAL than in FMOD.
Sound levels which were a perfect balance between music & game sound effects under FMOD instead had the music overpowering the *significantly* quieter sound effects when running under OpenAL.

addictgamer:
Also, it seems like changing the gamma setting has no effect until the game is restarted. I tested it against master and it's working fine there. This needs to be fixed.

addictgamer:
I'm encountering a very strange issue in OS X: after the first right click (or left click inside of the world view if in inventory mode), no further clicks will trigger grabbing an item or left-click-identifying objects in the world.

akien-mga:
> ```
#ifdef APPLE
    #include <al.h>
#else
    #include <AL/al.h>
#endif
```

Actually that would be better:
```
#ifdef APPLE
    #include <OpenAL/al.h>
#else
    #include <AL/al.h>
#endif
```

Same thing for OpenGL headers. Apple is weird sometimes :)

**Comment by ptitSeb** [comment]
I'll fix the MacOS X issue. I completly forgot to change the "SwapBuffer" for MacOS, that's why right click is not working.

For OpenAL/FMOD sound, I'm not sure. I have to recheck how the volume stuff is working for both API. But it may be the 3D sound that works slightly different. I've tried to match the model on both, but I may have missed something.

**Comment by addictgamer** [comment]
> Actually that would be better:

```
#ifdef APPLE
    #include <OpenAL/al.h>
#else
    #include <AL/al.h>
#endif
```

Ah, so it does work. Nice.


> I'll fix the MacOS X issue. I completly forgot to change the "SwapBuffer" for MacOS, that's why right click is not working.

> For OpenAL/FMOD sound, I'm not sure. I have to recheck how the volume stuff is working for both API. But it may be the 3D sound that works slightly different. I've tried to match the model on both, but I may have missed something.

Sounds good.

File: CMakeLists.txt (hunk 2 above)
Comment (ptitSeb) [inline CMakeLists.txt]: Yes, you are right. I think I have the LIB correctly set, but somehow forgot to change the INCLUDE part.

**Comment by addictgamer** [comment]
Looking good. OS X is building and running just fine now, everything should be working :+1:
Similarly, no issues in Linux.
(Altough, OS X is telling me the program has "Quit Unexpectedly" when I close Barony -- only when compiled with OpenAL)

Pending compilation and further testing under Windows.

**Comment by addictgamer** [comment]
It doesn't seem to wish to compile under Windows. I'm constantly receiving the following error:
```
Error	LNK2001	unresolved external symbol "void __cdecl closeOPENAL(void)" (?closeOPENAL@@YAXXZ)	editor	C:\programming\turning_wheel\barony\VS.2015\editor\init.obj	1	
```
For both editor & barony, in init.cpp.

Is there any reason why closeOPENAL is not declared in sound.hpp, and is instead forward declared in init.cpp?

**Comment by addictgamer** [comment]
Oh, I found what the issue was. closeOPENAL was declared `int` in sound.cpp, but `void` in init.cpp. Changing the declaration in init.cpp resolved that linker error.

However, new one:
`'openal_oggrelease': must return a value`
Adding a return statement at the end of said function solved that one too.

Once I fixed those two issues, it looks like everything works under Windows.

**Comment by ptitSeb** [comment]
Ah, damn, I'll fix that in the pull request.

**Comment by ptitSeb** [comment]
Is the macOS build still crashing at exit?

**Comment by ptitSeb** [comment]
I tested the GoG version (tagged 2.0.2 in the finder) on a mac and it was also telling me that Barony as Quitted unexpectly. That will make things difficult to fix.

**Comment by ptitSeb** [comment]
And I don't reproduce the issue with the OpenAL build. I think it maybe random, or semi-random issue.

**Comment by addictgamer** [comment]
Hmm. On this branch, I only saw that behavior with the OpenAL build.
What MacOS version are you testing on? I am running MacOS Sierra 10.12.3.

> Is the macOS build still crashing at exit?
Yes

**Comment by ptitSeb** [comment]
It was on Sierra too (not sure sub version). But I only saw it 1 time, with th eoriginal GoG version. I then launched 2 time the OpenAL version without issue. I haven't done much, just launch and quit from the menu.

**Comment by ptitSeb** [comment]
Yes, it's Sierra 10.12.3 too. Built with Xcode 8.2.1

**Comment by ptitSeb** [comment]
2017-02-01_17h11_19

I have this with the GoG version. I'll add a failsafe in the OpenAL version to avoid a similar segfault, just in case.

**Comment by ptitSeb** [comment]
Done. I hope it fixed the crash.

**Comment by addictgamer** [comment]
Doesn't look like that did it.

openal-crash

**Comment by ptitSeb** [comment]
Looks like the system cannot close OpenAL properly and launch an exception.
I tried to close sounds that could still be open... just in case (but I'm not confident it will fix that). Do you do something special on Mac to trigger this? You start game or you don't need to?

**Comment by addictgamer** [comment]
Nothing special,I just start the game.
Also, I don't know what happened, but that crash image I posted earlier isn't right? Here's the right one:

openal-crash-2

**Comment by ptitSeb** [comment]
Aaaahhh, so it's not about OpenAL, it's a about that sdl_mutex.

**Comment by addictgamer** [comment]
Ya. I think the other one may have been an earlier crash report or something, my bad.

**Comment by addictgamer** [comment]
In case it helps:

openal-crash-3

**Comment by ptitSeb** [comment]
Uh, the thread "openal" is still there? But how?! I'm supposed to call SDL_WaitThread to wait until this thread finished.

**Comment by ptitSeb** [comment]
I added more failsafe, and a warning in log.txt if something is wrong with the openal thread, if you can retry.

**Comment by addictgamer** [comment]
Ran it several times, seems like no more crash :)

**Comment by ptitSeb** [comment]
\o/

**Comment by addictgamer** [comment]
I'm going to wrap up testing today and hopefully we'll be good to merge!

**Comment by ptitSeb** [comment]
Ok good :)

**Comment by addictgamer** [comment]
All done, I think we're good to go.

Awesome work.

File: src/interface/interface.cpp [hunk 3]
```diff
@@ -145,7 +145,7 @@ bool loadInterfaceResources()
 	status_bmp = loadImage("images/system/StatusBar.png");
 	character_bmp = loadImage("images/system/CharacterSheet.png");
 	hunger_bmp = loadImage("images/system/Hunger.png");
-	minotaur_bmp = loadImage("images/system/minotaur.png");
+	minotaur_bmp = loadImage("images/sprites/minotaur.png"); // the file "images/system/minotaur.png" doesn't exist in current Data
```
addictgamer: Just noticed this change. That is correct, images/sprites/minotaur.png is a new resource. It should not be images/sprites/minotaur.png.

addictgamer:
Umm, so, after extensive testing, I've found out that boulder traps are not passable, as if there's an invisible wall there.

@ptitSeb I narrowed it down to the NEWLOOP code. If I go to commit 53baadc3d22e57ccc7c482cdcddb01250b7a21d6 and undefine NEWLOOP, boulder traps are passable once more.

src/interface/interface.cpp
ptitSeb: Yes, I needed the game to work, I haven't found that ressoyrce on the GoG data nor on the git source.
I can make a pull request to put it back to the correct value.

ptitSeb:
I'll recheck the loop. I must have remove some "z axis" (with map on x,y axis) test without noticing.

src/interface/interface.cpp
addictgamer: No, it's fine. I'll do it.

That resource is brand new, not available in the v2.0.4 release. It will be available in the next release. For now, to get the game to run, people will need to either create a dummy file or just load in a different file on their end -- as you have done.
Maybe I should make a note somewhere concerning new resources not yet available in the steam/gog release.

ptitSeb:
I'm a bit confused on why it does that.

in that `barony_clear(...)` function, the case where `my` is the player and `entity`, in the for`` loop at line 399 haven't really changed for boulderTrap, because `entity->flags[PASSABLE]` is true for bouldertrap (or minotaur trap) and so no much test is done.
I have indeed reproduce the behavour, but still scratching my head on this one.
Maybe the issue is with a test the other way, with `my` beeing the boulderTrap and `entity` beeing the player (but that part also seems to work, as trap seems to trigger?).

Still debugging...

ptitSeb:
One thing I have notice is that, in maps.cpp, line 2374: https://github.com/TurningWheel/Barony/blob/master/src/maps.cpp#L2374
a childEntity entity is created for the boulderTrap, but the `flags[PASSABLE] = TRUE` is set to the entity, not the childEntity here (entity already have that flag anyway). Is that correct, or should childEntity be used here?

ptitSeb:
I finally found the issu, and the fix is in the current pull request.

Issue was indeed with the 4 childEntity created for the boulderTrap. I assume these are the trigger for the trap. The issue was because those have a sizex/sizey of 0, and the NEWLOOP working on interval, the null size was not handled correctly.
I added a new test and it's now working fine.
I still think there is a typo in line 2374 of maps.cpp, and the `flags[PASSABLE] = TRUE` should be set on childEntity and not entity (but the size being 0, it doesn't matter much)

srathbunvl:
you're right, that's probably a typo. I had the same issue when I messed with the behavior for entities with size 0 bbox. Glad you located a fix however 👍 we'll review this in detail and pull it in when it's fully tested.

PR #54, opened 2017-02-08
Added some Hash map for Entities

ptitSeb:
The entity::uid is now private, with getUID() and setUID() accessor. getUID() is inlined, so read access should be at the same speed. setUID() on the other hand will update the map.
Also a small change to the way the minimap is drawn (should help with non-optimised drivers on low-end card).

### DIFF
--- a/src/actarrow.cpp
+++ b/src/actarrow.cpp
@@ -169,7 +169,7 @@ void actArrow(Entity* my)
 						if ( hit.entity->skill[0] != 1 && (hitstats->type < LICH || hitstats->type >= SHOPKEEPER) )
 						{
 							hit.entity->skill[0] = 2;
-							hit.entity->skill[1] = parent->uid;
+							hit.entity->skill[1] = parent->getUID();
 							hit.entity->fskill[2] = parent->x;
 							hit.entity->fskill[3] = parent->y;
 						}
@@ -193,7 +193,7 @@ void actArrow(Entity* my)
 											if ( hit.entity == entity )
 											{
 												entity->skill[0] = 2; // path state
-												entity->skill[1] = parent->uid;
+												entity->skill[1] = parent->getUID();
 												entity->fskill[2] = parent->x;
 												entity->fskill[3] = parent->y;
 											}
--- a/src/actarrowtrap.cpp
+++ b/src/actarrowtrap.cpp
@@ -83,7 +83,7 @@ void actArrowTrap(Entity* my)
 				{
 					Entity* entity = newEntity(166, 1, map.entities); // arrow
 					playSoundEntity(my, 239 + rand() % 3, 96);
-					entity->parent = my->uid;
+					entity->parent = my->getUID();
 					entity->x = my->x + x;
 					entity->y = my->y + y;
 					entity->z = my->z;
--- a/src/actbeartrap.cpp
+++ b/src/actbeartrap.cpp
@@ -77,7 +77,7 @@ void actBeartrap(Entity* my)
 	for ( node = map.entities->first; node != NULL; node = node->next )
 	{
 		Entity* entity = (Entity*)node->element;
-		if ( my->parent == entity->uid )
+		if ( my->parent == entity->getUID() )
 		{
 			continue;
 		}
--- a/src/actboulder.cpp
+++ b/src/actboulder.cpp
@@ -606,7 +606,7 @@ void actBoulderTrap(Entity* my)
 					if ( !map.tiles[OBSTACLELAYER + y * MAPLAYERS + x * MAPLAYERS * map.height] )
 					{
 						Entity* entity = newEntity(245, 1, map.entities); // boulder
-						entity->parent = my->uid;
+						entity->parent = my->getUID();
 						entity->x = (x << 4) + 8;
 						entity->y = (y << 4) + 8;
 						entity->z = -64;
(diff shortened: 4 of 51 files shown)


addictgamer:
Any particular reason behind changing UID from Uint32 to Sint32?

ptitSeb:
Yes, there are many case where the UID is -1, -2, -3 or -4. So to avoid some casting, I changed it to signed (plus I use a `if (uid<0)` test to swich between hash map and old method search.

srathbunvl:
Can you show some spots in the original code where the uid is treated like a signed int? UID should always be unsigned. If there are places in the code where it's treated as less than 0, those are mistakes. Also, it's better to have only one method (hash map) than to have both (linear and hash map).

ptitSeb:
Sure, here are all the places where an UID is set
```
/media/sda1/sources/git/barony$ grep setUID src/* -rn
src/actflame.cpp:60:            entity->setUID(0);
src/actflame.cpp:85:    entity->setUID(-3);
src/actgib.cpp:173:     entity->setUID(-3);
src/actmonster.cpp:120: entity->setUID(uid);
src/actplayer.cpp:2649:                                         entity2->setUID(-4);
src/actplayer.cpp:2653:                                         entity2->setUID(-3);
src/actplayer.cpp:2669:                                         entity2->setUID(-4);
src/actplayer.cpp:2673:                                         entity2->setUID(-3);
src/actsprite.cpp:109:  entity->setUID(-3);
src/actsprite.cpp:178:  entity->setUID(-3);
src/actsprite.cpp:248:  entity->setUID(-3);
src/acttorch.cpp:50:    entity->setUID(-3);
src/entity.cpp:136:void Entity::setUID(Sint32 new_uid) {
src/entity.hpp:65:      void setUID(Sint32 new_uid);
src/game.cpp:478:                                                                                               entity->setUID(-3);
src/game.cpp:1182:                                                                                              entity->setUID(-3);
src/magic/actmagic.cpp:2155:    entity->setUID(-3);
src/magic/actmagic.cpp:2206:            entity->setUID(-3);
src/maps.cpp:2008:                                              childEntity->setUID(-3);
src/monster_minotaur.cpp:802:                   entity->setUID(-3);
src/monster_minotaur.cpp:918:   childEntity->setUID(-3);
src/net.cpp:744:        entity->setUID((int)SDLNet_Read32(&net_packet->data[4])); // remember who I am
src/net.cpp:1111:                               entity2->setUID(entity->getUID());
src/net.cpp:1747:                       entity2->setUID(entity->getUID());
src/net.cpp:2024:                                               tempEntity->setUID(SDLNet_Read32(&net_packet->data[8 + 4 * (c - 2)]));
src/net.cpp:2028:                                               tempEntity->setUID(SDLNet_Read32(&net_packet->data[8 + 4 * (c - 1)]));
```
Just a simple grep (yeah, I'm a bit lazy).

My understanding is that negative values are used when not sure of the UID yet and/or for multiplayer clients

addictgamer:
Been doing some playtesting, it looks like all's working so far. Will do some more testing this weekend.

srathbunvl:
Sorry for the few days delay, had some downtime at work and doing some major upgrades to my desktop at home.

I remember what the negative UIDs are for, now... as you guessed, they are usually used by clients to mark entities that the server has/needs no knowledge about (eg flame particles) or they are used by the server to inform the client what "type" of object it is sending to the client. In other words, a definite hack... In general, all valid UIDs should be positive, and I used negative numbers in these cases because those equate to very large unsigned numbers. However, I was naive about the fact that this would add an additional cost to cast... therefore, the ideal solution would be to set uid to unsigned, and change these negative "magic numbers" to very large unsigned values (eg UINT_MAX-1, etc.) (or maybe a #define / const int is better?)

Also, just to reiterate on my last point, there should be only one search implementation (hash map) :)

ptitSeb:
I keep the Signed UID to differenciate between truly Unique ID (i.e. positive ones) with not Unique ID (the negative one, and also the 0 probably).
I can work on putting everything in the map (and also using unsigned instead of signed), but in that case, if there is a search for an not unique UID, chances are it will not found an entity and return NULL (because each time a "-4" for example is added, the map will only store 1, that will removed when any entity "-4" is destroyed).
I'll check tonight how will that work that way (unsigned with constant and all in the map).
I'll add the result to the pull request if it looks stable on my side.

ptitSeb:
Ok, done. Seems to works fine.

WALLOFJUSTICE:
How's the entity performance on this pull request? Looking forward to integrating this into my fork!

ptitSeb:
On the Pandora, that is very CPU limited (with it's 1GHz ARM CPU), you can clearly feel a difference, but I guess it will be less spectacular on x86 CPU.
I also plan to try do the same (implement some hash map) based on X/Y coordinate of the entity. It's quite some work, but I hope it will bring even more performances improvement.

addictgamer:
Apologies for the delays, I've been extraordinarily busy.

I think this is good to pull in, I slapped Baphy in the face without any issues.

addictgamer:
Also, for the record, use of auto is discouraged as per the coding styles guidelines.

addictgamer:
Oh, it looks like permissions were changed again, I didn't notice that until now. Rectified in master.

ptitSeb:
Sorry for the permission issue, I have to be more careful about that.

On the "auto" subject, I am not found of them, but I find them handy for iterator. I can replace auto with proper iterator definition if you prefer.

PR #60, opened 2017-04-25
[WIP] Dynamically generate resolutions in settings menu

pwhelan:
Here is a working implementation of dynamically querying and then showing resolutions in the settings menu. For the moment the biggest problem with the code is that it will leak memory since the list is initialized when the settings menu is shown. Another problem is that I do no sorting on the actual list of resolutions.

### DIFF
--- a/src/menu.cpp
+++ b/src/menu.cpp
@@ -4874,20 +4874,38 @@ void getResolutionList()
 	// display.
 	int numdisplays = SDL_GetNumVideoDisplays();
 	int nummodes = SDL_GetNumDisplayModes(0);
+	int im;
+	int c;
 	
 	printlog("display count: %d.\n", numdisplays);
 	printlog("display mode count: %d.\n", nummodes);
+	
+	for (im = 0; im < nummodes; im++)
+	{
+		SDL_DisplayMode mode;
+		SDL_GetDisplayMode(0, im, &mode);
+		// resolutions below 960x600 are not supported
+		if (mode.w < 960 || mode.h < 600)
+		{
+			nummodes--;
+		}
+	}
 	
 	resolutions.num = nummodes;
 	resolutions.modes = (struct resolution *)
 		calloc(nummodes, sizeof(struct resolution));
 	
-	for (int im = 0; im < nummodes; im++)
+	for (im = 0, c = 0; im < nummodes; im++)
 	{
 		SDL_DisplayMode mode;
 		SDL_GetDisplayMode(0, im, &mode);
-		resolutions.modes[im].width = mode.w;
-		resolutions.modes[im].height = mode.h;
+		if (mode.w < 960 || mode.h < 600)
+		{
+			continue;
+		}
+		resolutions.modes[c].width = mode.w;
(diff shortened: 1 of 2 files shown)


pwhelan:
@ptitSeb any chance of you testing this on Pandora? It *should* just work (thanks to the magic of SDL2).

addictgamer:
Awesome, can't wait to take a look at it later this week.

At a cursory glance, I think it should not display resolutions below a certain size. If the resolution is not high enough, the game isn't playable, and 960x600 is the lowest resolution I've been testing all of my work against, I cannot guarantee any sort of quality for resolutions much lower than that.

ptitSeb:
@pwhelan : sure, I'll try to test that this weekend on Pandora.

pwhelan:
> If the resolution is not high enough, the game isn't really playable, and 960x600 is the lowest resolution I've been testing against

I can simply filter out resolutions below a certain width and/or height.

addictgamer:
0x0 and repeating resolutions

I'm getting 0x0 as a resolution (4 times apparently), and several other resolutions are duplicated. (Selecting any of the duplicates highlights all the rest, in a similar manner to how 0x0 is behaving in the screenshot)

WillKirkby:
Willing to bet those duplicate entries are coming from SDL returning multiple entries where a given resolution has multiple available pixel modes and/or refresh rates.

pwhelan:
To make ordering as well as filtering out duplicates I was thinking of using an ordered linked list. Thoughts?

pwhelan:
Is it fair game to use sec/list.cop?

SheridanR:
src/list.cpp works just fine. I'd do an insertion sort to organize the resolutions and drop ones that are unneeded / broken. Just make sure to free the data when the game shuts down or the window closes.

lheckemann:
FWIW I've [modified](https://github.com/lheckemann/Barony/commit/34397d749093e5de5697954a42b76f2ed4286b51) this to use a standard linked list rather than an ad-hoc vector. I don't know why the original code doesn't use the standard library much, so maybe there's an obstacle to using this that I don't know of, but this seems like a simpler option and reduces the amount of code necessary so I'm using it at least for myself.

lheckemann:
One issue I see with this method in general is that it queries available fullscreen modes, which may exclude some usable windowed ones.

addictgamer:
>  I don't know why the original code doesn't use the standard library much

For the record, the original code was all C. If the codebase looks like a C program that compiles with a C++ compiler, that's pretty much what's going on.

lheckemann:
That explains a lot!

addictgamer:
I don't oppose using the C++ standard library, but I note I want cross-platform compatibility. std::rand, for example, doesn't abide by that, so we use prng in places where the server and client have to get the same result (so the game can be played between different OS!).

In general, I prefer new code and contributions to lean towards C++ rather than sticking with the legacy C style. And parts of the old codebase have been roughly ported to C++. It's a bit of a mess.

lheckemann:
Alright. In that case, I'll go ahead and PR my version which fixes the duplicated and 0x0 resolutions.

PR #62, opened 2017-04-27
Add new items, fixes to editor

WALLOFJUSTICE:
Editor items sprites now dynamically change depending on what you choose.

Added hover tooltip for sprite selection with name and sprite index.

Added items.
-Rebalanced Steel Shield of Resistance
-Added scaling hp/mana regen, as well as magic resistance.
-Reformatted armor breaking code to handle items of varying likelihood to break - please review the order in which things happen, I believe they're the same as the old method.

-If possible ignore any VS2015/git files that were pushed.. they shouldn't be there


src/entity.cpp
```diff
@@ -1693,26 +1693,54 @@ void Entity::handleEffects(Stat* myStats)
...
 	int healring = 0;
+	int healthRegenInterval = 0;
 	if ( myStats->ring != NULL )
 	{
 		if ( myStats->ring->type == RING_REGENERATION )
 		{
 			if ( myStats->ring->beatitude >= 0 )
 			{
-				healring = 1;
+				healring++;
```
addictgamer: For the record, we prefer pre-increment to post-increment, unless the behavior of post-increment is what is semantically desired.

PR #64, opened 2017-05-02
Abstract over data dir

lheckemann:
This allows things like passing a different data directory on the
command line, which can be useful for packaging.

### DIFF
--- a/src/buttons.cpp
+++ b/src/buttons.cpp
@@ -313,7 +313,7 @@ void buttonOpen(button_t* my)
 	button->focused = 1;
 
 	// file list
-	if ( (dir = opendir("maps/")) != NULL )
+	if ( (dir = openDataDir("maps/")) != NULL )
 	{
 		while ( (ent = readdir(dir)) != NULL )
 		{
@@ -340,7 +340,7 @@ void buttonOpen(button_t* my)
 			d_names[c] = (char*) malloc(sizeof(char) * FILENAME_MAX);
 		}
 		c = 0;
-		if ( (dir = opendir("maps/")) != NULL )
+		if ( (dir = openDataDir("maps/")) != NULL )
 		{
 			while ( (ent = readdir(dir)) != NULL )
 			{
@@ -484,7 +484,7 @@ void buttonSaveAs(button_t* my)
 	button->focused = 1;
 
 	// file list
-	if ( (dir = opendir("maps/")) != NULL )
+	if ( (dir = openDataDir("maps/")) != NULL )
 	{
 		while ( (ent = readdir(dir)) != NULL )
 		{
@@ -511,7 +511,7 @@ void buttonSaveAs(button_t* my)
 			d_names[c] = (char*) malloc(sizeof(char) * FILENAME_MAX);
 		}
 		c = 0;
-		if ( (dir = opendir("maps/")) != NULL )
+		if ( (dir = openDataDir("maps/")) != NULL )
 		{
 			while ( (ent = readdir(dir)) != NULL )
 			{
--- a/src/editor.cpp
+++ b/src/editor.cpp
@@ -703,6 +703,7 @@ void redo()
 void processCommandLine(int argc, char** argv)
 {
 	int c = 0;
+	strcpy(datadir, "./");
 	if ( argc > 1 )
 	{
 		for ( c = 1; c < argc; c++ )
@@ -714,9 +715,14 @@ void processCommandLine(int argc, char** argv)
 					strcpy(maptoload, argv[c] + 5);
 					loadingmap = true;
 				}
+				else if (!strncmp(argv[c], "-datadir=", 9))
+				{
+					strcpy(datadir, argv[c] + 9);
+				}
 			}
 		}
 	}
+	printlog("Data path is %s", datadir);
 }
 
 /*-------------------------------------------------------------------------------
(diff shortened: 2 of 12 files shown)


lheckemann:
WIP: doesn't handle fonts yet

lheckemann:
Fixed

addictgamer:
I'm sorry I haven't had a chance to look over your pull requests yet, I've been flooded with business lately.
Bearing that in mind, how does this play with OSX? I set it up so that the game grabs its current running directory and then changes its running directory to ../Content/
The editor does similar, instead changing to Barony.app's Content folder. I'd imagine this should work just fine regardless, and I will test it when I get a chance.

lheckemann:
No worries!

Yes, this should still work correctly with OSX — I didn't remove any of the `chdir`s and the application still searches the working directory by default.

I forgot about the editor though, I think this might actually break it completely (because datadir is uninitialised in its case). I'm not sure though, I haven't tried it.

addictgamer:
The editor indeed doesn't work:
```
[23-52-29] error: unable to locate language file: 'lang/en.txt'
```

Running the game, there's a few things I noticed:
https://pastebin.com/DHPQ9ZEL
```
[23-39-13] Opening .//books/
[23-39-13] Opening .//books/Worker's Journal, Entry 4.txt
[23-39-13] Could not open './/books/Worker's Journal, Entry 4.txt': Not a directory
```
What's up with those errors, it looks like it's trying to open every file in the books directory as a folder?
After a big block of that repeating for most every book file, it looks like it then proceeds to open the books just fine.

`[23-39-00] Opening .//lang/en.txt` Not dealbreaker, but can anything be done about that .// prepending everything?

lheckemann:
On the "not a directory" errors: yes, I believe the code is searching the whole directory tree, and trying to open everything in the directory as a directory for that purpose. That's not new, only the error checking is. Given that the books are in a flat hierarchy, it might make sense to remove the tree search and just have it go through all the files in there.

About the `.//`, would simply removing the log messages do? It has no effect on the behaviour and I only really added those messages to work out which stuff was loading correctly. Any other option would require some special-casing which doesn't really seem necessary.

lheckemann:
That's the editor fixed.

addictgamer:
Cool, editor's working fine. However, it seems the code is still looking for steam_appid.txt in the binary's directory rather than the specified data dir. I don't remember, was the code that looked for that hidden away in the Steamworks SDK? It looks like it is.

addictgamer:
Also, it looks like music doesn't work if the music files are in the data dir but not the binary's dir?
It just repeats the following in the terminal:
[19-51-17] Can't play music.

No sound either.

addictgamer:
Oh, and for the record, the code runs just fine on Mac, generally speaking. Game starts up as normal and so does the editor.

addictgamer:
> That's not new, only the error checking is. Given that the books are in a flat hierarchy, it might make sense to remove the tree search and just have it go through all the files in there. Alternatively, we could get rid of the error messages.

Ah. Nah, I'm fine with it as is for now. At some point, it might be a good idea to make it only try to open directories, not files.

> About the .//, would simply removing the log messages do? It has no effect on the behaviour and I only really added those messages to work out which stuff was loading correctly. Any other option would require some special-casing which doesn't really seem necessary.

I agree about special casing being unnecessary, I'm fine with removing the logging.

lheckemann:
A quick grep yielded no references to steam_appid.txt in the code. However, I don't think the steam case matters much here anyway as when it's run from steam it will always be running with its containing directory as its working directory.

For the music and sound, I presume it isn't working because I didn't apply the change for FMOD (which is because I'm unable to build and test that). It works with OpenAL.

For the books again, it doesn't actually search recursively after all — just tries opening everything as a directory to check if it is a directory. Using `stat` would probably be more suitable for this, so I've made that change.

addictgamer:
> A quick grep yielded no references to steam_appid.txt in the code. However, I don't think the steam case matters much here anyway as when it's run from steam it will always be running with its containing directory as its working directory.

Fair point.

> For the books again, it doesn't actually search recursively after all — just tries opening everything as a directory to check if it is a directory. Using stat would probably be more suitable for this, so I've made that change.

Sounds good

> For the music and sound, I presume it isn't working because I didn't apply the change for FMOD (which is because I'm unable to build and test that). It works with OpenAL.

Ok. Then I'll have to look into it later this week.

addictgamer:
> > For the music and sound, I presume it isn't working because I didn't apply the change for FMOD (which is because I'm unable to build and test that). It works with OpenAL.
> Ok. Then I'll have to look into it later this week.

No, I'll accept this PR as is. Applying this change to FMOD wouldn't be worth my time, since those builds aren't using the data-dir abstractions.

addictgamer:
I just noticed there's some large merge conflicts in loadMap() with the barony-next branch.

lheckemann:
I can probably port it to barony-next, but is there anywhere I can get data files for testing it with?

addictgamer:
I can assemble some dummy files.

addictgamer:
[baronynext-dummy.zip](https://github.com/TurningWheel/Barony/files/1009518/baronynext-dummy.zip)

@lheckemann this should do it. Run it on top of the files from the latest steam release (v2.0.5).

lheckemann:
Thanks!

WALLOFJUSTICE:
Should be easy to sort out.. just eyeballed the differences now. Let me know if there's any trouble because I haven't touched any original directory stuff

PR #67, opened 2017-05-13
Query SDL for resolutions

lheckemann:
Based on @pwhelan's #60, but fixes the multiple-entry and 0x0 bugs.

### DIFF
--- a/src/menu.cpp
+++ b/src/menu.cpp
@@ -9,6 +9,7 @@
 
 -------------------------------------------------------------------------------*/
 
+#include <list>
 #include "main.hpp"
 #include "game.hpp"
 #include "stat.hpp"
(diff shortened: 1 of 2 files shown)


lheckemann:
Um... I think you broke it @addictgamer  :p

addictgamer:
Ya, meant to push only one commit. Instead it pushed the whole branch. My bad.

lheckemann:
```
git checkout resolution-auto
git fetch
git reset --hard @{u}
git cherry-pick 3b45538 
git push
```

Should fix it with only the one commit I presume you intended

addictgamer:
Much better :)

addictgamer:
Finished testing in Mac and Windows, looks like it works.
I dislike how it's throwing away all of the resolutions I use on my desktop, but this can be improved later.

Perhaps the stock resolutions we've been shipping with so far should arbitrarily get sorted into the list.

PR #70, opened 2017-05-18
New Tileset MonsterCurve Added, Editor Tooltips

WALLOFJUSTICE:
Editor now has button for sprite name hovering tooltip when nothing is selected, enabled by default, CTRL H to toggle.

TurnL TurnR raise/lower z of player under noclip, shift halves movement.

MonsterCurve includes 2 new tilesets, still to balance.

Updated lang file in repo.


src/editor.cpp
```diff
@@ -3297,7 +3297,10 @@ int main(int argc, char** argv)
 				case 75:	strcpy(action,"TROLL"); break;
 				default:	strcpy(action,"STATIC"); break;
 			}*/
-			if ( palette[mousey + mousex * yres] >= 0 )
+
+			int numsprites = (int)sizeof(spriteEditorNameStrings) / sizeof(spriteEditorNameStrings[0]);
```
addictgamer: For the record, we prefer C++-style casts over C-style casts.

addictgamer:
For items, can it be changed to include the item name, if that was set?
Same for monsters, show the monster's name if set.

PR #71, opened 2017-05-19
Data dir next

lheckemann:
#64 ported to barony-next

### DIFF
--- a/src/buttons.cpp
+++ b/src/buttons.cpp
@@ -343,7 +343,7 @@ void buttonOpen(button_t* my)
 	button->focused = 1;
 
 	// file list
-	if ( (dir = opendir("maps/")) != NULL )
+	if ( (dir = openDataDir("maps/")) != NULL )
 	{
 		while ( (ent = readdir(dir)) != NULL )
 		{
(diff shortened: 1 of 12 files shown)


addictgamer:
So, it seems there's an issue with books. When I tried to pick up the journal in the starting level, the game crashed in items.cpp line 367:
```snprintf(&tempstr[c], 1024 - c, language[1007], books[appearance % numbooks]->name);```

Inspecting the game state, I found out numbooks is 0.
Looking at log.txt, I found:
```
[02-56-58] compiling books...
[02-56-58] Warning: discoveredbooks->first and last do not exist. No books in /books/ directory?
```
My datad-r was ./ and the books directory definitely was populated.

lheckemann:
Yep, I've been trying to work that one out. I'll get back to you once I've fixed it.

lheckemann:
Turns out `==` has higher precedence than `&` >_>

Fixed.

addictgamer:
Yep, that fixed it.

I did notice if datadir isn't passed in with a trailing slash, it fails to load the random player names files.

PR #77, opened 2017-05-29
Remove SDLMain.h and SDLMain.m

HiPhish:
Those files were needed for version 1 of SDL on OS X, but SDL 2 does not need them. They are not used anywhere in the code, so it is safe to remove them.


addictgamer:
Hi @HiPhish,

I'm on vacation until August, when I return home I'll pull out my Mac and look both into this and #78

addictgamer:
Ya, looks like leftover junk.

PR #87, opened 2017-06-13
Use C++ list for book discovery

lheckemann:
Less code, more efficient sorting. Yay!

### DIFF
--- a/src/book.cpp
+++ b/src/book.cpp
@@ -23,7 +23,6 @@
 
 book_t** books = NULL;
 int numbooks = 0;
-list_t* discoveredbooks = NULL;
 
 //book_t *book_space_ninjas = NULL;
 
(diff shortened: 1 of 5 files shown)


src/book.cpp
```diff
@@ -42,72 +41,44 @@ int getBook(char* booktitle)
...
-	node_t* node;
 	string_t* name = NULL;
-	bool unsorted;
 	int i = 0;
 
 	//TODO: Read the books/ inventory for all *.txt files.
 	//TODO: Then create a book for each file there and add it to a books array.
-	discoveredbooks = directoryContents("books/");
-	if (discoveredbooks)
+	auto discoveredbooks = directoryContents("books/");
```
gavrilyuc: need changed `auto` to `std::list<std::string>`

lheckemann: Barony uses C++11 features like `auto` in other places, this should work if your compiler settings are correct I believe.

addictgamer: > Barony uses C++11 features like auto

For the record, the coding style guidelines prefer avoidance of auto.
(It seems Barony is in violation of many things...)

PR #92, opened 2017-07-09
Add power crystal, FPS slider, editor bear trap

WALLOFJUSTICE:
++ fix crossbow rapid fire exploit
++ fix multiplayer games resuming from save game getting de-synced
++ update lang file


src/maps.cpp
```diff
@@ -3578,6 +3578,35 @@ void assignActions(map_t* map)
 
 				break;
 			}
+			// set beartrap
+			case 107:
+			{
+				entity->skill[0] = 1; // so everything knows I'm a chair
```
addictgamer: haha

PR #95, opened 2017-07-29
Factoring language memory management freeing code and

devnexen:
couple of resource leaks fixes.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -339,10 +339,12 @@ int initApp(char* title, int fullscreen)
 			if ( c == 0 )
 			{
 				printlog("sprite 0 cannot be NULL!\n");
+				fclose(fp);
 				return 7;
 			}
 		}
 	}
+	fclose(fp);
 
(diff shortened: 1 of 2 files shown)


devnexen:
Just realised that the good branch was barony-next, hopefully this and #97 will pass now :-)

src/init.cpp
```diff
@@ -383,10 +385,12 @@ int initApp(char* title, int fullscreen)
 			if ( c == 0 )
 			{
 				printlog("model 0 cannot be NULL!\n");
+				fclose(fp);
 				return 12;
 			}
 		}
 	}
+	fclose(fp);
```
addictgamer: This one is redundant and causes a crash, fp is freed several lines lower. One of them must go:
```
	fclose(fp);
	if ( !softwaremode )
	{
		generatePolyModels(0, nummodels);
	}
	fclose(fp);
```

addictgamer:
Looks good

PR #97, opened 2017-07-31
Build test under BSD, those systems need extra lib (e.g execinfo)

devnexen:
for backtrace calls.
The compiler got confused with the below statement which is seen
as a full boolean expression which is compared to NULL, so make the
initial intent clearer.


addictgamer:
Looks good

PR #105, opened 2017-08-07
Refactor 'int multiplayer' into 'NetworkType localPlayerNetworkType'

crkellen:
I didn't branch when I started the refactor on NetworkType, so the first couple of commits are explained here:
---
+ Created enum PlayerClass to remove ambiguity

* Cleaned up and reorganized charclass.cpp to better align with style conventions
* Changed the name of enum Status to ItemStatus for more clarity
* Changed the name of enum Category to ItemCategory for more clarity

- Removed the stray Config.hpp in VS-includes

---

Main portion of the changes are focused on converting the global variable 'int multiplayer' to 'NetworkType localPlayerNetworkType'. Extra changes were made along the way to better align with style conventions. Some of those outliers are listed below:

+ Added enum NetworkType to replace #defines for SINGLE, SERVER, CLIENT, etc.
+ Added general comments for clarity or definition of functions
+ Added comments in a few places to translate what language[x] means as a string
+ Added 'TODOR:' comments in multiple locations. TODOR means TODO REFACTOR, and are places where refactoring needs to be done. This will help avoid issues where a strange location is not checked while refactoring a certain system.

* Refactored getSaveGameType() to be clearer and return a NetworkType
* Renamed 'int mul' to 'NetworkType saveGameNetworkType' for clarity
* Refactored openFailedConnectionWindow() to be clearer and accept a NetworkType
* Renamed 'int multiplayerselect' to 'NetworkType selectedNetworkType' for clarity
* Fixed general issues with ints being used as bools (multiplayer was used as a bool/assigned ints)


src/charclass.cpp
```diff
@@ -27,19 +27,35 @@
 
 void initClass(int player)
 {
+    // TODO: Dynamic class list loading/saving from/to file
+    enum class classes_t
```
addictgamer: I think a cleaner solution here be to use static consts.
E.g:

```
static const int BARBARIAN = 0;
static const int WARRIOR = 1;
...
```

crkellen: So that the values are still ints?

addictgamer: Nnnnot exactly. The problem isn't so much what the values are as much as the casting and extra typing it introduces.

Regardless, it's good you brought this up. Magic numbers are A Bad Thing and should be eliminated.

crkellen: It does require the extra typing of PlayerClass::BARBARIAN, however I added the "class" modifier to prevent scoping issues. The casting is an issue, and I believe I made a few TODOR:'s to look into dealing with a few of those.

There's many instances of ints being used as bools, and eliminating them is something I'd like to continue to look into. Creating an enum was one such method, but only the first step.

PR #106, opened 2017-08-07
Kobold/Golem/Cockatrice animated/itemized/statted

WALLOFJUSTICE:
* Quick create multiplayer lobby for debug, ctrl + L (Lobby) at start menu
* Quick join multiplayer lobby for debug, ctrl + M (Multiplayer) at start menu for above lobby.
* Magic code integer division converts to double first, never really impacted most spells but come across this bug when testing my new spells.
* Fix filthy rich steam achievement not granting for clients.
* Smashed the lang files together


src/monster_cockatrice.cpp
```diff
@@ -19,24 +19,28 @@
 #include "net.hpp"
 #include "collision.hpp"
 #include "player.hpp"
+#include "magic\magic.hpp"
```
addictgamer: Don't use `\` for filepaths, it's not cross-platform. Use `/` for filepaths.

src/items.cpp
```diff
@@ -2439,3 +2439,145 @@ void createCustomInventory(Stat* stats, int itemLimit)
 		}
 	}
 }
+
+node_t* itemNodeInInventory(Stat* myStats, Item* itemToFind, Category cat)
```
addictgamer: This either needs a check to ensure that the function is not run when itemToFind is `nullptr`, or, better yet, make myStats and itemToFind references rather than pointers.

addictgamer: Or I guess not for this specific case, seeing as `itemToFind` was changed to a non-pointer type. Regardless, in general, this applies. If a function needs an argument, said argument usually should be a reference.

src/actmonster.cpp
```diff
@@ -4939,3 +4943,89 @@ void handleMonsterSpecialAttack(Entity* my, Stat* myStats, Entity* target, doubl
 	return;
 }
 
+void getTargetsAroundEntity(Entity* my, Entity* originalTarget, double distToFind, real_t angleToSearch, int searchType, list_t** list)
```
addictgamer: As a rule of thumb, anything taking a my argument probably belongs as a member of that class.

src/entity.cpp
```diff
@@ -4005,6 +4011,29 @@ void Entity::attack(int pose, int charge, Entity* target)
...
+						// crystal golem special attack increase chance for armor to break if hit. (25-33%)
+						// special attack only degrades armor if primary target.
+						else if ( pose == MONSTER_POSE_GOLEM_SMASH && target == nullptr )
+						{
+							if ( isWeakArmor )
+							{
+								// 66% chance to be deselected from degrading.
+								if ( rand() % 3 > 0 )
+								{
+									armor = NULL;
```
addictgamer: Reminder that `nullptr` is preferred over `NULL`.

Issue #104, opened 2017-08-05
[2.0.5][Windows][Multiplayer] Client has difficulties "bumping" another player


crkellen:
**ISSUE:**
Players can "bump" each other by right clicking. This displays their name, current HP/Max, etc. Clients in a multiplayer game are often unable to preform this action, but the Host can freely.

**EXPECTED RESULT:**
When a Client player right-clicks another player, it should preform the "bump" action.

**STEPS TO REPRODUCE:**
1. Host a multiplayer game.
2. Invite or have friends join through Steam.
3. Start the game.
4. Right-click the other players, note any difficulties in doing so. The text should appear in the bottom right.
5. Have someone else Host a multiplayer game.
6. Join through Steam.
7. Start the game.
8. Attempt to right-click other players, note any difficulties in doing so. The text should appear in the bottom right.

**ADDITIONAL INFO:**
By the time we reached the underworld, the "bump" action was completely impossible for the Client with me to do. I could still do it freely. Often times he has to be slightly inside of my entity (pushing me), to get it to trigger.

SheridanR:
I believe it always works for clients as long as you click on the player's head. It has nothing to do with proximity or collision.

The same issue occurs for talking to followers or shopkeepers. It only works for clients if you click on their head.

Clicking on limbs is supposed to do a lookup on the client to get the character's head uid, which iirc is the "actual" object-click that is sent to the server. But something along that pipeline isn't working right.

crkellen:
Yep. @WALLOFJUSTICE explained that only the head exists, and everything else is simulated locally.
I'll take a look at it sometime.

crkellen:
The issue with this was due to the Client being unable to select entities that were not the head.
This was because the entities were not in the 'map_entities' map that contains all entities.
This was fixed by changing a return to a continue statement.

addictgamer:
Re-opening this since I want it merged into an off-shoot of master and shipped in a live build.

crkellen:
As requested, there is now a new PR based on v2.0.7 branch.

addictgamer:
Merged in #218 into v2.0.7-branch

PR #111, opened 2017-08-12
Fix issue #104 - Client does not add Limb Entities to map

crkellen:
This is the fix for the issue that Client's could only Right-Click on player's and shopkeepers heads. This was because the limb entities were not added to 'map_entities'. This was fixed by changing the 'return' to a 'continue'.


addictgamer:
Good work

addictgamer:
I'd like to get this out in path 2.0.7.

I'm not sure what the best way to go about doing it is, but let's try this: could I get another PR for this based off of `2.0.7-branch?`

crkellen:
Sure I can handle that in a little bit here.
null

PR #112, opened 2017-08-13
Tidy up of attack animation code

WALLOFJUSTICE:
Attack animation should be consistent in multiplayer now. 
Refactor one of the main attack functions to be called on the entity itself rather than just the limb, still todo the other one.
Potentially need to adjust all humanoid bent limb checks to make sure they're always bent


src/entity.cpp
```diff
@@ -6107,16 +6111,16 @@ bool Entity::hasRangedWeapon() const
 	return false;
 }
 
-void Entity::handleWeaponArmAttack(Entity* my)
+void Entity::handleWeaponArmAttack(Entity* weaponarm)
 {
-	if ( my == nullptr )
+	if ( this == nullptr || weaponarm == nullptr )
```
addictgamer: C++ Standard: 8.3.2/4:
>Note: in particular, a null reference cannot exist in a well-defined program, because the only way
to create such a reference would be to bind it to the “object” obtained by indirection through a null pointer,
which causes undefined behavior.

The entity variable should be verified it's not nullptr before calling any functions on it, not checked for nullptr inside of the class's member functions.

PR #113, opened 2017-08-13
[QoL] [Feature] Better HP/MP bars, HP/MP displayed in Character Sheet

crkellen:
This is a Quality of Life (QoL) feature addition, meant to improve the clarity of how much health and mana you have, along with displaying maximum HP/MP in the character sheet.
Attached below is a picture of the before and after. These additions are small, simply adding information to the character sheet and expanding it, and adding a separator between the top of the colored bar, and the bottom of the bar's title text.

New and Old

### DIFF
--- a/lang/en.txt
+++ b/lang/en.txt
@@ -2960,4 +2960,6 @@ dominate the %s into your service!#
 2472 You feel magic weaken against you.#
 2473 You feel magic bounce off you.#
 
+2499 HP: %d/%d   MP: %d/%d#
+
 2500 end#
--- a/lang/en.txt
+++ b/lang/en.txt
@@ -576,7 +576,6 @@ thou makest use of them.#
 1203 INT:          %2d -> #
 1204 PER:          %2d -> #
 1205 CHR:          %2d -> #
-2499 HP: %d/%d   MP: %d/%d#
 
 
 # skill levels
--- a/lang/en.txt
+++ b/lang/en.txt
@@ -2960,7 +2960,6 @@ dominate the %s into your service!#
 2472 You feel magic weaken against you.#
 2473 You feel magic bounce off you.#
 
-2499 HP: %d/%d#
-2500 MP: %d/%d#
+2499 HP: %d/%d MP: %d/%d#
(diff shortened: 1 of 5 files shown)


crkellen:
Due to the back and forth testing which UI will be best, I will create a new pull request for this, once the idea has been finalized.

lang/en.txt
```diff
@@ -2960,4 +2960,6 @@ dominate the %s into your service!#
 2472 You feel magic weaken against you.#
 2473 You feel magic bounce off you.#
 
-2499 end#
+2499 HP: %d/%d MP: %d/%d#
```
lheckemann: Might make sense to put this closer to the rest of the character sheet strings, e.g. at position 1206 which isn't occupied.

lheckemann:
Also, I don't see any reason to create a new PR. Just replace the commits in this one, that way the discussion on the PR stays in one place :)

crkellen:
I was instructed to place it at the end, to prevent conflicts. It may be moved at a later time, possibly if merged.
Additionally, the Style Guidelines I believe said to use real spaces, not Tabs. I may be wrong and will have to check.
null

lheckemann:
Oh, I see.

[Style guidelines](https://docs.google.com/document/d/1Wx-tkiNORweFd6htDOn88QG5i4ecyulMGVdXQhxhQnw/edit#) say:

>Use real tabs that equal 4 spaces.

crkellen:
I have strived to follow those guidelines, and other than using Allman (over One True), my settings were the same. However, if the Style Guidelines are outdated, or there is any other conflict. It isn't much trouble for me to update it. 
I would prefer to stay consistent with others, as long as I do not sacrifice readability or robustness.
null

lheckemann:
As it says, *real* tabs with a width of 4 spaces, but *not* actual spaces.

crkellen:
I'll update accordingly most likely tomorrow.
null

crkellen:
This actually should have been closed. There's a better implementation now. Also it was partially added by #200.

Issue #114, opened 2017-08-14
[2.0.5][Multiplayer][Client] Fountains do not stop animating for Client


crkellen:
**ISSUE:**
When fountains are right-clicked, they get used, causing their animation to stop. This is not the case on Client players, who will always see the fountains animation.

**EXPECTED RESULT:**
When someone, Host or Client, uses a fountain, it should stop it's animation.

**STEPS TO REPRODUCE:**
1. Be a Client
2. Find a fountain
3. Drink from it, notice the animation does not stop.

**ADDITIONAL INFO:**
It does not matter who drinks from the fountain. It will always animate.

addictgamer:
Resolved in #204

PR #118, opened 2017-08-15
Fix Issue #114 - Client fountains do not stop animating

crkellen:
This is a fix for Issue #114.
Clients were not being sync'd the changes to the fountain's status.
A simple call to 'serverUpdateEntitySkill()' solved that issue.

An optional change would be changing usage of 'my->skill[3]' to 'my->skill[2]' now that 'my->skill[2]' is unused.


src/actfountain.cpp
```diff
@@ -215,6 +215,7 @@ void actFountain(Entity* my)
 					}
 					messagePlayer(i, language[474]);
 					my->skill[0] = 0; //Dry up fountain.
+                    serverUpdateEntitySkill(my, my->skill[0]);
```
addictgamer: Out of curiosity, is there a `Sint32&` defined for this in class Entity in Entity.hpp?
E.g. `Sint32& fountainStatus`?

addictgamer: Also, it looks like the formatting for this line still needs to be fixed?

crkellen:
I'll handle this today when I can.
null

addictgamer:
> I'll handle this today when I can.

Ok.
If there's an Sint32& reference, don't switch to it yet; we added some of those in barony-next and they don't exist in master.

Speaking of which, can you rebase this PR against master?

crkellen:
Yes, I will rebase this, along with any others you want against master. Just let me know which ones you want rebased.
null

addictgamer:
Merged into master in #204

Issue #102, opened 2017-08-04
[2.0.5][Multiplayer][Client] Client will continue to see Host on fire after they step in water


crkellen:
**ISSUE:**
When a Host is set on fire, other Clients can see an effect of sparks coming off of that player. When a Host is on fire and steps into water (starts swimming), both the effect of sparks and the status of being "On Fire" are removed. However, for the Clients, they continue to see that the Hosthas the spark effect.

**EXPECTED RESULT:**
When a player loses a status, such as being on fire, the effect corresponding to that status should also be removed.

**STEPS TO REPRODUCE:**
Requires water, so the Swamp is the testing ground.
1. In a multiplayer Steam game, have a method of lighting someone on fire, or being lit on fire, without dying. (Goblins with Fire rods work well).
2. Have the host be lit on fire, then submerge themselves in water. The effect will remain for players other than the host (verified with two clients).
3. Have a client be set on fire, then submerge themselves in water. (The effect is not visible on the clients #100 but normally, it would be removed).

**ADDITIONAL INFO:**
I haven't fully tested this one, and I will update this issue as necessary. I know for now this is an issue with clients, I believe I have been a host and watched a client submerge themselves and lose the spark effect, but I have not verified yet. I cannot remember if the effect persists while in water, or only after exiting.

It's very likely that similar to #100 the host sees the correct behavior, but every client does not.
I have been told that it eventually went out. I was also told that while I was invisible (cloak of invisibility), and not on fire anymore, the effect still lasted for a long time. It is very possible that this issue is actually:
**"On Fire" effect on another player does not disappear if it is removed before timing out (clients only)**

crkellen:
Updating the title and OP of this, as through more testing, the issue has been narrowed down.
The Host does not send an update when they remove the fire from themselves when going into water.
A pull request will follow shortly.

addictgamer:
Resolved in #203

PR #119, opened 2017-08-15
Fix Issue #102 - Host "On Fire" status does not disappear

crkellen:
This is a fix for #102.
The issue is caused by if statement checks that say if the player is a Client, update your status. These statements were removed from both stepping in water, and drinking from a fountain.

Cases that need to be verified:
- ~~Player using Cure Ailment Spell~~ - Verified that it works on Clients viewing the host. Cannot fully verify due to #120 
- ~~Player using Cure Ailment Potion~~ - Verified that it works on Clients viewing the host. Cannot fully verify due to #120
- ~~Player's "On Fire" Status timing out~~ - Verified that Clients will continue to see the effects on the Host after it times out. **Commit https://github.com/TurningWheel/Barony/pull/119/commits/3340353d96b45155e86a2fb7ada6306846183987**

Just a thought, I will look into this tomorrow, but perhaps the function:
~~'serverUpdateEntityFlag()' is the cause of the issues, not the check.~~
The issue is that the Server handles removing the "On Fire" status on Clients, but not setting Clients to be on fire.. There is a
`if ( multiplayer != CLIENT )` check before every time `flags[BURNING] = false` then the check after which never executes:
`if ( PLAYER_NUM > 0 )` which would call the 'serverUpdateEntityFlag()'. Removing the check fixes this, although in general, the net code structure needs a redesign at some point, as it is ambiguous how well it aligns with proper Server Authoritative structure.


addictgamer:
What's the status of this? Is the fix completed and ready to be merged?

crkellen:
Yes this is completed. I can verify as such in 9 hours or so.
null

addictgamer:
Ok, just wasn't sure on the status of it.

crkellen:
After reviewing the post I can see why. I'll edit it as well when I can.
null

addictgamer:
Merged in #203

addictgamer:
Resolved in #203

Issue #100, opened 2017-08-04
[2.0.5][Multiplayer][Client] Client does not see sparks coming from them while on fire


crkellen:
**ISSUE:**
When a client player is "on fire" they do not have anything to tell them that they are on fire, other than the continual health loss. Client players can see sparks coming off of other players when those other players are on fire, but do not see the effects in first person on themselves.

**EXPECTED RESULT:**
When on a player is on fire, they should be able to see the effect of sparks coming off of their body, similar to how they see other players on fire, but in first person.

**STEPS TO REPRODUCE:**
FOR CLIENT:
1. Have someone host a multiplayer game.
2. Join that game through Steam.
3. Play the game until you have a way to set yourself on fire, or be set on fire.
4. Have the other player(s) set themselves on fire, or be set on fire.
FOR HOST:
1. Host a multiplayer game.
2. Have someone join your game through Steam.
3. Play the game until you have a way to set yourself on fire, or be set on fire.
4. Have the other player(s) set themselves on fire, or be set on fire.

**ADDITIONAL INFO:**
I was unaware of this for the longest time, as I am always the host. A lot of my friends continually die because they have no idea they're on fire, and don't notice the health loss. A suggestion comes from this as well, that it's pretty difficult to tell when you're taking damage, or that you have low health.

crkellen:
I have found the solution to this, and will implement and test tomorrow. Expect a pull request tomorrow when it's fully tested.

addictgamer:
Resolved in #202

PR #122, opened 2017-08-15
Fix Issue #100 - Client cannot see themselves "On Fire"

crkellen:
This is a fix for #100.
The issue was caused by issues similar and related to #102 and #119. Adding a call to 'serverUpdateEntityFlag()' is all that was missing in this case. Every case of a Client being set on fire works.


addictgamer:
Merged #202 into Master.

Issue #116, opened 2017-08-15
[2.0.5] When typing keyboard inputs still function


crkellen:
**ISSUE:**
When a player, singleplayer or otherwise, presses enter and begins typing, all inputs are still valid. This means that when a player looks around, left or right clicks, or presses WASD, all functionality is still the same. However, when hitting WASD to move, they player does not move, but still animates, causing them to bounce slightly in place.

**EXPECTED RESULT:**
Player's keyboard inputs should be nullified/frozen for the duration of the chat window being open. So when WASD is pressed, the player does not animate in an attempt to move.

**STEPS TO REPRODUCE:**
1. Start a game.
2. Type in chat, press W, A, S, and D keys.

**ADDITIONAL INFO:**
A slight annoyance, it doesn't affect gameplay as far as I can tell, but should still be fixed.

addictgamer:
Resolved in #206

PR #123, opened 2017-08-15
Fix Issue #116 - Head Bobbing when typing

crkellen:
This is a fix for #116.
This issue was fixed by adding a check of '!command' to the head bob code, similar to how movement forces are not calculated if '!command'.


crkellen:
Thanks!I'll update accordingly soon.
null

addictgamer:
Merged in #206

Issue #120, opened 2017-08-15
[2.0.5][Multiplayer][Client] Cure Ailment does not work on Client


crkellen:
This issue is the same as #42.
**ISSUE:**
When Clients are on fire, and they attempt to either drink a potion of Cure Ailment, or cast the spell, it will display the message that they are refreshed, but it will not remove the fire. They will continue to take damage, cloaks will continue to burn. They are still on fire.

**EXPECTED RESULT:**
When using a Spell or Potion of Cure Ailment on the Client, while on fire, it should remove the "On Fire" effect.

**STEPS TO REPRODUCE:**
1. Be a Client in a multiplayer game
2. Have a wizard shoot you with fireball
3. Attempt to cast Cure Ailment or drink a Potion of Cure Ailment
4. The effect will not be removed.

**ADDITIONAL INFO:**
~~This has only been seen so far in branch 'barony-next'. I will attempt to test it tomorrow on the live Steam version.~~ This issue is also present on live. The Potion of Cure Ailment does correctly remove the flame effects on the Client (after fixing #100), but the spell does not. The effects are removed, but the spell still is affecting them, causing damage and burning cloaks.

crkellen:
An additional note, the Potion of Cure Ailment does correctly remove the flame effects on the Client (after fixing #100), but the spell does not.

crkellen:
Resolved in #207.

Issue #42, opened 2016-12-26
Casting cure ailment spell as a client does not put out flames on caster


addictgamer:

addictgamer:
Same issue with potions.

addictgamer:
Resolved among various PRs, including #207.

PR #132, opened 2017-08-15
Fix Issue #120 - Cure Ailment does not remove Fire on Clients

crkellen:
This is a fix for #120 and #42.
The issue was the spell of Cure Ailment was checking only the local player (host), for the burning flag. This was fixed to check for the player that is actually casting the spell.


crkellen:
Resolved in #207.

Issue #128, opened 2017-08-15
[2.0.5][Multiplayer] Being hit with Locking displays wrong message


crkellen:
**ISSUE:**
If a Player is hit with the spell of Locking, it displays the message "You open the door!" on the one who was hit.

**EXPECTED RESULT:**
If a Player is hit with the spell of Locking, it should display "No telling what it did..."

**ADDITIONAL INFO:**
This assumes that Friendly Fire is turned ON.

addictgamer:
Resolved in #211

PR #133, opened 2017-08-15
Fix Issue #128 - Hit by Locking Spell displays wrong message

crkellen:
This is a fix for #128.
The issue was just that the language value being used to display the message was off by 1.


addictgamer:
Merged in #211

Issue #126, opened 2017-08-15
[2.0.5][Multiplayer][Client] Being hit with Slow doesn't work on Client


crkellen:
**ISSUE:**
If a Client is hit with the spell of Slow, it does not slow them.

**EXPECTED RESULT:**
If a Client is hit with the spell of Slow, it should slow them.

**ADDITIONAL INFO:**
This assumes that Friendly Fire is turned ON.

addictgamer:
Resolved in #209

PR #134, opened 2017-08-15
Fix Issue #126 - Client is not affected by Slow Spell

crkellen:
This is a fix for #126.
The issue was that the Client was not being synced the fact that their status effects had been changed. Appropriate calls were added to fix that, to both the Slow spell, and the Cold spell, which applies the Slow effect.


addictgamer:
Merged in #209

Issue #127, opened 2017-08-15
[2.0.5][Multiplayer][Client] Being hit with Confuse does't work on Client


crkellen:
**ISSUE:**
If a Client is hit with the spell of Confuse, it does not confuse them.

**EXPECTED RESULT:**
If a Client is hit with the spell of Confuse, it should confuse them.

**ADDITIONAL INFO:**
This assumes that Friendly Fire is turned ON.

addictgamer:
Resolved in #210

PR #135, opened 2017-08-15
Fix Issue #127 - Client is not affected by Confuse Spell

crkellen:
This is a fix for #127.
The issue was that the Client was not being synced the fact that their status effects had been changed. The appropriate call was added to the Confuse spell to fix that.


addictgamer:
Merged in #210

Issue #130, opened 2017-08-15
[2.0.5] Players are not notified when Slow effect wears off


crkellen:
**ISSUE:**
If a Player is slowed, when the effect times out, the Player is not notified.

**EXPECTED RESULT:**
If a Player is slowed, when the effect times out, they Player should receive a message notifying them they are no longer slowed.

**ADDITIONAL INFO:**
This is mainly to be consistent. If a Player is standing still, they may not realize they aren't slowed anymore. In general, user feedback is important.

addictgamer:
Resolved in #213

PR #136, opened 2017-08-15
Fix Issue #130 - Notify Player when they are no longer Slowed

crkellen:
This is a fix for #130.
The issue was there was a missing case statement for `EFF_SLOW` which meant nothing would be done once the Player was no longer slowed. Adding a case that displays the same message as `EFF_FAST` "You return to your normal speed." is a good fix.


addictgamer:
Merged in #213

Issue #48, opened 2017-01-17
Remove Curse Spell does not work as intended for clients in multiplayer


addictgamer:
When playing as a client in multiplayer, the remove curse spell does not function as intended.
It *should* pop open the Remove Curse GUI. Instead, when a client receives the `RCUR` packet, it runs the code for the remove curse scroll (which, indeed, also triggers the `RCUR` packet).

Prescribed remedy:
Create a new packet for this situation.
When that packet is received, pop open the Remove Curse GUI.

addictgamer:
Fixed in #320

PR #137, opened 2017-08-15
Fix Issue #48 - Client Remove Curse Spell does not open GUI

crkellen:
This is a fix for #48.
The issue was there was never a new packet implemented on the Client side to open the GUI. One was implemented for Identify ('IDEN') but there was one lacking for Remove Curse.

- Function 'closeIdentifyGUI()' was added to keep consistent behavior.


crkellen:
This has been replaced by #177. As such, it is no longer needed. This pull request **_should not_** be valued over the other.

Issue #129, opened 2017-08-15
[2.0.5] Player can cast spells while asleep


crkellen:
**ISSUE:**
If a Player is asleep, they can still cast any spell.

**EXPECTED RESULT:**
If a Player is asleep, they should be unable to do anything.

**ADDITIONAL INFO:**
Although you can cast spells, you can't aim, so wherever you were looking last, is where it will shoot.

addictgamer:
Resolved in #212

PR #138, opened 2017-08-16
Fix Issue #129 - Entities can cast spells while Asleep

crkellen:
This is a fix for #129.
The issue was there was a missing check in 'castSpellInit()' for seeing if the Entity was Asleep. There was already one for being Paralyzed, but not one for `EFF_ASLEEP`.


addictgamer:
Merged in #212

Issue #43, opened 2016-12-26
Casting spells while swimming still drains MP


addictgamer:
Probably shouldn't do that. Should just instantly fail.

PR #142, opened 2017-08-16
Fix Issue #43 - Prevent MP loss when casting spell while swimming

crkellen:
This is a fix for #43.
The issue was twofold:
- There was no check if the Caster was swimming in 'castSpellInit()', which meant the spell was always being processed. A check for swimming was added to 'castSpellInit()' to solve this issue.
- The check inside of 'castSpell()' was after MP had already been drained. Moving it before this point solves the issue.


src/magic/castSpell.cpp
```diff
@@ -110,6 +110,39 @@ void castSpellInit(Uint32 caster_uid, spell_t* spell)
...
+        if ( stat->shoes != nullptr )
+        {
+            if ( stat->shoes->type == IRON_BOOTS_WATERWALKING )
+            {
+                bIsCasterWaterWalking = true;
+            }
+        }
+
+        // If both cases are false, the Caster could potentially be swimming
+        if ( bIsCasterLevitating != true && bIsCasterWaterWalking != true )
```
lheckemann: `!= true` is redundant on bools :)

src/magic/castSpell.cpp
```diff
@@ -110,6 +110,39 @@ void castSpellInit(Uint32 caster_uid, spell_t* spell)
 		return;
 	}
 
+    // Check to make sure the Caster is not swimming
```
lheckemann: There's a lot of conditions here which are duplicated (not only in here but in `actplayer.cpp`, `acthudweapon.cpp`, which should probably be factored out to reduce the duplication and the amount of code that needs to be changed should additional conditions for swimming be added.

crkellen:
I'll review this one further today or tomorrow.
null

crkellen:
Please see pull request #171. It is a _**replacement**_ for this one, and also a fix for #143.
This pull request _**should not**_ be valued over the other one.

Issue #141, opened 2017-08-16
[2.0.5][Multiplayer] Casting Magic Missile with 100 Spellcasting still drains MP


crkellen:
**ISSUE:**
If a Player has level 100 in the proficiency Spellcasting, they are supposed to be able to cast the spell Magic Missile without any mana cost. In Singleplayer this works. In Multiplayer, even with no others in the game, this does not work.

**EXPECTED RESULT:**
If a Player has level 100 in the proficiency Spellcasting, they should be able to cast the spell Magic Missile without any mana cost.

**STEPS TO REPRODUCE:**
1. Start up a Singleplayer game with 100 Spellcasting and a book of Magic Missile
2. Cast Magic Missile, your mana will NOT decrease.
3. Disconnect back to the Main Menu.
4. Host a Multiplayer game.
5. Start the game with or without players
6. Cast Magic Missile, your mana WILL decrease.

**ADDITIONAL INFO:**
The server should not be sending packets to itself, so it is likely missing a check somewhere for client or local player.

addictgamer:
`[2.0.5]`? This should only be present in barony-next...

addictgamer:
Will need to confirm the bug myself but I'm going to preemptively place the bug label.

addictgamer:
Actually, let's do that. `bug-to-confirm`
Will help me keep the two sorted.

addictgamer:
I can't believe I forgot, but it looks like I already confirmed this:

`//TODO: Fix issue #141: Spellcasting capstone doesn't work in multiplayer.`

I need a break.

crkellen:
This was merged by https://github.com/TurningWheel/Barony/commit/4b16b1cd0e45de255197a66183eeab899b50bfe8. Closing.

PR #145, opened 2017-08-17
Fix Issue #141 - Multiplayer Magic Missile cost at 100 proficiency

crkellen:
This is a fix for #141.
The issue was in Multiplayer there was not an additional check to see if the Caster had 100 proficiency in Spellcasting while casting Magic Missile. There needed to be an additional one here as 'getCostOfSpell()' does not have access to the Player proficiency.


crkellen:
Somehow, this branch got insanely mangled. I believe it was when I was attempting to create a build for v2.0.5 with my fixes. Deleting this in order to recreate it.

Issue #151, opened 2017-08-17
[2.0.5] Pools of Blood can hide Items on the ground


crkellen:
**ISSUE:**
When an Entity dies or spawns a Pool of Blood, small Items such as gloves, pickaxes, quarterstaffs, etc, can become hidden inside of the Pool of Blood, often obscuring it's existence from the Player.

**EXPECTED RESULT:**
Pools of Blood should be unable to fully hide Items from sight.

**ADDITIONAL INFO:**
This is referenced in #44.

addictgamer:
Resolved in #214

PR #152, opened 2017-08-17
Fix Issue #151 - Pools of Blood can hide Items

crkellen:
This is a fix for #151.
The issue was the height of spawning the pools of blood was high enough that it would fully cover a small item such as a Ring. Along with this, the extra, unneeded calculation added to the end of the pool of blood's Z position was removed as there is no reason to have variation in the height of the pool of blood.

- Modified height of pools of blood spawned on Entity death
- Modified height of pools of blood spawned from Bleed status.


addictgamer:
Merged in #214

Issue #125, opened 2017-08-15
[2.0.5] Spell of Opening just opens the door, doesnt unlock


crkellen:
**ISSUE:**
When using the Spell of Opening on a locked door, it only opens the door, it doesn't unlock it. Causing the door to be stuck in a open state. Doors in the open state cannot be attacked, or have spells casted on them.

**EXPECTED RESULT:**
When casting the Spell of Opening on a locked door, the door should unlock, not open.

**ADDITIONAL INFO:**

addictgamer:
Resolved in #208

PR #154, opened 2017-08-17
Fix Issue #125 - Spell of Opening does not unlock doors

crkellen:
This is a fix for #125.
The issue was there was no change to the actual lock status of the door being hit, the Spell of Opening would simply do as it's name implies and "open" the door. This however, puts the door in a intangible state, as it is no longer a physical Entity.

Additionally, I opted to remove the effect of opening the door altogether, as I believe it would be better to leave that up to the Player's discretion. By letting them decide whether or not to open the door, prevents issues where a mob runs in when the Player is not ready (swapping items/spells, or otherwise).


SheridanR:
Good fix for the bug... the spell should open the door though. if you don't want to let monsters in, don't cast the spell.

crkellen:
I actually reverted that locally. I realized soon after that it opens metal gates. So I fixed it and made it open still. Guessing never pushed the changes. Will do when I get a chance.
null

crkellen:
I went ahead and remedied this issue, originally, I didn't realize the Spell of Opening could actually open Iron gates and the like. I simply assumed it was just a counter spell to Locking, and thus "Opening" was really more just "Unlocking".

It now unlocks, and opens.

crkellen:
I went ahead and extended the formatting update to the rest of the Spell of Opening as well. Therefore IF statements have the spaces between braces and such as it should be.

src/magic/actmagic.cpp
```diff
@@ -1909,6 +1909,8 @@ void actMagicMissile(Entity* my)   //TODO: Verify this function.
 					{
 						if (hit.entity->behavior == &actDoor)
 						{
+                            playSoundEntity(hit.entity, 91, 64);
+                            hit.entity->skill[5] = 0; // Unlock the door.
```
addictgamer: Out of curiosity, is there any `Sint32& ...` defined for this in Entity.hpp? Like `Sint32& doorLocked`?
The chest has a bunch of these sorts of defines, like `Sint32& chestOpener` IIRC

crkellen: As far as I can see, there is not. It is however possible that `chest_opener(skill[5])` could be used, however for now, I'll keep it unmodified.

addictgamer:
Merged in #208

PR #155, opened 2017-08-17
couple of small code corrections :

devnexen:
- few small mem leaks on error return paths.
- possible overflow when number of players == MAXPLAYERS.
- a pointer freed too early leading to use after free.
- a boolean function not returning.


crkellen:
It would be useful if your changes were made using separate commits.
So that this pull request would be 5 commits, with each commit having a small description of the issue it fixes.

The first one in actmonster.cpp would then be a commit of that one addition `return true;` and the commit message could be "Fixed boolean function not returning"

devnexen:
Ok did not dare first regarding the few changes it implied. Done now !

crkellen:
Much better! Very easy to see what change does what, and able to review separately.

Issue #82, opened 2017-06-05
Possible to push boulder onto self


lheckemann:
It has occasionally happened to me that I was killed by pushing a boulder. I don't know the exact conditions required to reproduce this, but it's a very annoying bug — as I see it, this should not be possible as the boulder is always being pushed away from the player pushing it.

SheridanR:
Highly agree. Not sure why I never thought of this solution...

crkellen:
I just wanted to comment that since this isn't closed, and I have lost multiple runs because of this, I'm going to look into fixing it.

lheckemann:
I can't seem to reproduce this issue anymore. Is it fixed in spite of @crkellen 's PRs not having been merged?

lheckemann:
Never mind, I was just pushing it wrong.

crkellen:
I was wondering the same. I can't check the latest code right now, but was there any changes to boulders done manually without the PR? I put off updating this because I hadn't experienced the issue.
null

crkellen:
I can update this sometime this week.
null

PR #159, opened 2017-08-18
Fix Issue #82 - Pushed boulder can collide with pusher

crkellen:
This is a fix for #82.
The issue is twofold:
1. Boulders pushing movement is calculated expected a perfect resolution. It uses truncation (floor()) in it's calculations, which cause it to be fixed to grid positions (only moves in increments of 16). This is only a problem when the boulder collides with something like another boulder (or other Entity), because the boulder will be somewhere in between the increment of 16. When a push happens after that, the floor() calculation causes the boulder to jump back to it's correct position (starting position). The player may be standing within that starting position, which causes the boulder to collide and hurt the one pushing.

2. Calculations for collisions with boulders is done using 'entityInsideEntity()'. This is a simple AABB bounding box collision check, which allows for a collision to occur from the opposite direction than the boulder is rolling. The solution is to implement 'entityInFrontOfEntity()' which, for each direction, avoids the possibility of colliding in the direction opposite of movement.

So, the collision is completely fixed, however there are still a few things left to sort out with boulders:
- Clients do not have accurate positions of boulders (constantly vibrating if not against a wall/entity)
- Boulders have a size that is 1 unit less than their actual radius. This allows for them to be partially inside of another Entity, such as another boulder, which causes a visual issue of the boulder jumping back when the floor() calculations are done.

These will be solved in future pull requests.


crkellen:
I agree, I guess I didn't consider that after finally figuring out the issue. All fixed now!

Issue #124, opened 2017-08-15
[2.0.5][Balance][Multiplayer] When reviving players should start with full HP/MP


crkellen:
**ISSUE:**
When a player dies, but the remaining part makes it to the next floor, they revive without their max health or mana. Loss of all of their gear, plus experience, causes this to contribute heavily to the snowball effect of constant player death.

**EXPECTED RESULT:**
When a player is revived, they should start with their full HP/MP.

**ADDITIONAL INFO:**
Usually what happens is whoever survives gets all of the XP and becomes a carry, while the rest of the players continue to die repeatedly without much hope to recover.

danielnelson:
I'm very sad to see the pull request was closed.  While death ending a run is fine in single player, I feel that multiplayer could be much improved if there was a way to make the game less punishing.  As it is, I just abandon the game and restart on first player death.

Perhaps we could have some options regarding player death penalties?  Personally, I would prefer if it was possible for players to keep their inventories.

crkellen:
The snowballing effect of a single player's death causes them to be forever behind in experience, loot, and fun.
I too believe it to a critical flaw in the game's design, when playing multiplayer. I believe a community vote, or more discussion would be a good step in the right direction.
null

SheridanR:
Loot and experience are both easily shared and there is no loss of experience, attributes, or skill points on death. Given that this is a roguelike, perhaps the penalty for dying should actually be greater than it is. HP is easily recovered (especially at later stages) so ultimately it's only a minor deterrent.

danielnelson:
There is always the opportunity costs of being dead, which are rather steep in my experience, but the biggest issue for me is that you usually cannot carry all of the items of the dead player, and even if you can it's time consuming which breaks the flow of the game and it's a problem to dispense the items on the next level before being mugged.

> Given that this is a roguelike, perhaps the penalty for dying should actually be greater than it is.

I guess it could be game over for them, but while very roguelike that doesn't make for a great multiplayer experience.  No one wants to sit around without playing for too long.

I think it is something that is player and group specific, everyone I have played with loses interest in replaying the first 3 levels over and over, and wants to move on to more forgiving games.  These are players who have no problem dealing with other fairly difficult games such as vagante.  Maybe those of us that are not as good can have the option of a easier experience?

SheridanR:
[...]

PR #164, opened 2017-08-19
Fix Issue #124 - Revive with full HP/MP

crkellen:
This is a fix for #124.
The issue was Host players were getting their HP and MP set to MAX / 2. This was changed to be just MAX.
Additionally, Clients did not receive any MP at all. Whatever they had when they died, was what they kept. To fix this, a new line was added to set it to MAX as well.


SheridanR:
Can't remember how mp was meant to be behave but the player reviving at half hp isn't an accident.

crkellen:
What is the reason for Players starting at half HP? Having that as a feature only seems to add to the multiplayer snowballing issue.

lheckemann:
I'm not here to guess at everyone's original intentions, but I think the snowballing effect is desirable: death is something to avoid very carefully, and certainly shouldn't be a routine occurrence from a gameplay perspective IMHO. It's already pretty unbalanced in that it allows surviving without enough food by simply taking it in turns to die, and increasing HP on revive just exacerbates that.

addictgamer:
HP and MP should both be half the max when a player revives.
If client's don't receive any MP, that's a bug that should be fixed to match the host's/singleplayer behavior.

crkellen:
I am still a bit confused as to why that is the intended behavior, but regardless, I can solve the issue tomorrow, as it is simply a single missing line in the clientsided code.
null

addictgamer:
Thanks.

I'm going to close this pull request then.

PR #166, opened 2017-08-20
Stalags, monsterAddNearbyItemToInventory + Automaton Recycling

WALLOFJUSTICE:


src/entity.cpp
```diff
@@ -7353,6 +7354,8 @@ void Entity::checkGroundForItems()
 		case GOATMAN:
 			//Goatman boss picks up items too.
 			checkBetterEquipment(myStats);
```
addictgamer: oops, goatman has no more break.

Issue #149, opened 2017-08-17
[2.0.5][Collision] Life Saving Amulet does not protect against boulders


crkellen:
**ISSUE:**
If a Player is wearing a Life Saving Amulet and dies to a boulder trap, they will be revived, only to be damaged again, often resulting in two deaths in quick succession, rendering the Life Saving Amulet useless against boulder traps.

**EXPECTED RESULT:**
When hit by a boulder while wearing a Life Saving Amulet, if they would die, they will be revived, and wont take further damage.

**ADDITIONAL INFO:**
This is a pretty easy fix. Add a check to the boulder collision with Players, if they are wearing a Life Saving Amulet, then destroy the boulder.

WALLOFJUSTICE:
Addressed in 41d55c71, finally

PR #167, opened 2017-08-21
Fix Issue #149 - Amulet of Life Saving does not protect against boulders

crkellen:
This is a fix for #149.
The issue was the existing code for being hit by a boulder did not take into account Amulets of Life Saving, and therefore, was useless to protect against them. The new implementation checks to see if the Entity is a Player, and is wearing an Amulet of Life Saving, if so, act as if the Player survived the hit, by destroying the boulder.

~~Note: This code will have to be duplicated for the additional function added in #159.~~
**Note:** This code will have to be re-implemented for the newest changes in #159. I will do that after #159 is merged.


crkellen:
I am rarely satisfied with my solutions, but have been instructed to not refactor as much as possible until the release of the new update.
There are many refactorings I have in mind, but until the release, duplications (following how the code currently exists) is the method I took.
I will review this further today or tomorrow.
null

crkellen:
After reviewing, I can say that I agree with your statements. When looking for how life saving amulets worked, it surprised me how it was implemented. I could potentially look into refactoring how death is handled at a later point. For now, I think it's best to simply implement it like this, although it'll have to be re-implemented for the new changes in #159.

WALLOFJUSTICE:
Did the changes in 41d55c71 to be merged, closing this PR

Issue #156, opened 2017-08-17
[2.0.5][Steam] Achievements can be earned while Cheats are turned on


crkellen:
**ISSUE:**
Steam Achievements can still be earned while cheats are turned on, allow Players to cheat to get achievements easier.

**EXPECTED RESULT:**
When the cheats flag is set, Steam Achievements should be disabled.

**ADDITIONAL INFO:**
This is a pretty easy change, so I'll probably make it later today.

PR #168, opened 2017-08-21
Fix Issue #156 - Disable Steam Achievements if Cheats are enabled

crkellen:
This is a fix for #156.
The issue was there was no checks implemented to see if Cheats were enabled. To fix this, a check was added to 'steamAchievementClient()' to prevent the sending of a packet to the client, and another check was placed in 'steamAchievement()' to prevent the actual granting of an achievement.

I cannot presently test this, however, it should work.


crkellen:
This has been updated with all of the changes requested.
It now does the following:
- [x] Displays a warning message when hovering over the box to enable cheats
- [x] Displays a warning message near the version string when cheats are enabled
- [x] Displays a warning message near the version string when achievements have been disabled for the session
- [x] Disables achievements for the session if a game is started while cheats are enabled.
- [x] Disables achievements for the session if a save is loaded where cheats were used.
- [x] Saves the usage of cheats into the save data to prevent save scumming to use cheats and get achievements
- [x] All code that actually preforms logic (other than handling save data) is surrounded in `#ifdef STEAMWORKS` blocks

Do note that this modifies the save data, and therefore is incompatible with saves prior to it.

src/main.hpp
```diff
@@ -714,3 +714,5 @@ bool dataPathExists(const char *const);
 bool completePath(char *dest, const char * const path);
 
 static const Uint32 cacheLimit = 8096;
+
+extern bool areCheatsDisabledForSession;
```
lheckemann: Shouldn't this be *achievements* rather than cheats?

lheckemann:
👎  on the "All code that actually preforms logic (other than handling save data) is surrounded in #ifdef STEAMWORKS blocks". There's no good reason for this and it makes the differences between differently compiled copies of the game bigger.

Other than that, LGTM (just the code, haven't tested it).

crkellen:
Yeah, I mixed words there by accident. I'll have to fix that.
Not sure why you don't want the ifdefs,as everything else Steam related is already #ifdef'd out.
null

lheckemann:
It's unnecessary extra code, which contributes to the mental load of reading it. I think it makes a lot more sense to only have the ifdef that makes `steamAchievement` a no-op when appropriate and just rely on that abstraction. In addition to making the code easier to make sense of it would make it less complicated to add another achievement mechanism for instance. This is similarly hard to maintain as the sound systems — rather than abstracting over them, openal code was added in loads of places that shouldn't care about which sound system is being used, just that they have a sound system. In this case we have two "achievement systems" — steam, and nothing. Any code that isn't implementing one of these shouldn't have to care which one is being used.

In summary: I'd prefer *removing* the ifdefs for achievements everywhere except within `steamAchievement`.

crkellen:
Okay, that sounds fair. Unless @addictgamer says otherwise, I'll make the changes likely Sunday.
null

PR #169, opened 2017-08-25
Linkage issue fix, especially isMeleeWeapon is inlined

devnexen:
monster_goatman cant's access its symbol.
isRangedWeapon thus not necessary is more a "preventive" fix.

### DIFF
--- a/src/items.cpp
+++ b/src/items.cpp
@@ -2587,7 +2587,7 @@ node_t* getMeleeWeaponItemNodeInInventory(Stat* myStats)
 	return nullptr;
 }
 
-bool inline isRangedWeapon(const Item& item)
+bool isRangedWeapon(const Item& item)
 {
 	switch ( item.type )
 	{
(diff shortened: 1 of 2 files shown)


devnexen:
It is a change to take with a grain of salt maybe those inlined function is just a first step of a more significant following change thus I might miss the big picture but ... just in case ;-)

addictgamer:
I don't understand, what was the linkage issue?

devnexen:
It does not compile for me, when it comes to monster_goatman.cpp.o object (which contains a call to isMeleeWeapon).

devnexen:
screenshot_2017-08-26_05-54-40

In case I was messing up with my own barony-next branch I just cloned a fresh one from this repo and got same linkage error as you can see but again if it s a part of a bigger planned change not comitted yet but then I can drop this PR no problems.

addictgamer:
Will have to look into this more, but I believe one of the other (more recent?) PRs addresses this already.

devnexen:
Yes I think @crkellen chose to put the implementation in the headers approach instead. Whatever is best ;-)

crkellen:
I'll review this after work today. I don't remember making a PR for any changes to inlining.
null

devnexen:
Sorry I confused you with @lheckemann :-) this is the PR in question https://github.com/TurningWheel/Barony/pull/179

lheckemann:
I prefer this one over mine (didn't see it when I was making mine!). Inline functions are confusing.

PR #172, opened 2017-08-27
Automaton recycle tweaks, potion throw animation for goatmen

WALLOFJUSTICE:
Throw animation needs a tidy up/it's own new windups but I threw something together heaps quick to expand upon.


src/monster_automaton.cpp
```diff
@@ -1131,7 +1235,15 @@ void Entity::automatonRecycleItem()
 	{
 		return;
 	}
+	
+	if ( this->monsterSpecialState == 0 )
```
addictgamer: All of this `this` is redundant, usually shouldn't be used.
Should phase out use of `this` in member functions and just access the members directly.

src/entity.cpp
```diff
@@ -7610,16 +7610,24 @@ void Entity::monsterAddNearbyItemToInventory(Stat* myStats, int rangeToFind, int
 				{
 					//Drop that item out of the monster's inventory, and add this item to the monster's inventory.
 					Item* itemToDrop = static_cast<Item*>(replaceInventoryItem->element);
-					if ( itemToDrop && myStats->type != AUTOMATON ) // Automatons are hoarders.
+					if ( itemToDrop )
 					{
-						dropItemMonster(itemToDrop, this, myStats, itemToDrop->count);
+						if ( !(myStats->type == AUTOMATON && list_Size(&myStats->inventory) < maxInventoryItems) )
```
addictgamer: If there's no else statement, shouldn't this be combined with the previous if statement?

addictgamer:
So the potion drink animation currently plays even when the goatman doesn't drink it, huh?

Issue #173, opened 2017-08-27
[2.0.5] Bows/Slings do not fire consistently


crkellen:
**ISSUE:**
Bows/Slings do not fire consistently while holding down left-click, and sometimes when just clicking left-click. Crossbows work as expected, consistently.

**EXPECTED RESULT:**
Holding down Left-Click with a Bow/Sling equipped should fire consistently the same as a Crossbow. However, there is still the bow draw animation.

**STEPS TO REPRODUCE:**
1. Start a new game as a Warrior or Rogue.
2. Equip the Shortbow
3. Hold down Left-Click
4. Shots will fire inconsistently. Continue to hold it down. Don't move mouse, then later move mouse. Try moving. It's all very very inconsistent.

**ADDITIONAL INFO:**
This seems to be because of the reliance on the sound clip playing. However, I haven't looked at it in-depth yet.

addictgamer:
Might need reevaluation after #86

PR #86, opened 2017-06-13
Barony next user data

lheckemann:
Same as #85 but for barony-next, I've resolved all the conflicts.

PR #174, opened 2017-08-28
Fix Issue #173 - Fix Bow Sustained Firing

crkellen:
This is a fix for #173.
The issue is that with OpenAL, undefined behavior can happen which is not present with FMOD. This undefined behavior is the program believing that the clip is still playing, or has started, despite it only playing once. Likely, this is not an issue with OpenAL, but rather, the game code is intertwined with the sound code, and the pointer is never cleared, which allows a previously ended sound clip to continue to exist until the next time it is assigned.

Pointers should not be kept alive to old objects. Undefined behavior is happening because until the bow is fired again, this pointer continues to exist with it's old data, and continues to try and process 'OPENAL_Channel_IsPlaying()'

Likely, this will not cause any issues with FMOD, and will actually be better. You feel free to test and correct me if I am wrong.


addictgamer:
May need review with #86 in mind.

src/acthudweapon.cpp
```diff
@@ -463,17 +463,18 @@ void actHudWeapon(Entity* my)
...
 								{
-									if (bowFire)
+									if ( bowFire )
 									{
 										bowFire = false;
+										bowDrawingSound = nullptr;
 										players[clientnum]->entity->attack(0, 0, nullptr);
 										HUDWEAPON_MOVEX = 3;
-										throwGimpTimer = TICKS_PER_SECOND / 4;
+										throwGimpTimer = 12; // Was TICKS_PER_SECOND / 4
```
lheckemann: Any particular reason for this change? Having it defined as "a quarter of a second" seems better than having it defined as "12 ticks", which would change firing behaviour if the tick rate is changed.

lheckemann:
Should be fine, I think.

crkellen:
I believe the reason for the change was because every frame it would be calculating the exact same value, as the framerate couldn't change (#define'd). I can revert it later today.
null

crkellen:
It could cause issues, but it shouldn't, since the actual issue is that the pointer is never cleared, allowing for loads of undefined behavior to be possible.
FMOD seems to either handle those issues on their end, or doesn't care about them.
null

lheckemann:
>I believe the reason for the change was because every frame it would be calculating the exact same value

The compiler can work that out just fine, don't worry about it.

PR #176, opened 2017-09-03
+ New editor map flags to change ceiling tile, disable monster/item spawns

WALLOFJUSTICE:
+ Refire single-boulder traps to editor
* Update insectoid movelimbs template to new style


addictgamer:
Looks like slings say "You are hit by an arrow!"
Is the insectoid supposed to drop the acid spellbook?

addictgamer:
Acid spray should probably bust doors too.

src/entity.cpp
```diff
@@ -7887,3 +7858,56 @@ Item** Entity::shouldMonsterEquipThisArmor(const Item& item) const
 	}
 }
 
+void Entity::degradeArmor(Stat* hitstats, Item* armor, int armornum)
+{
+	int playerhit = -1;
+
+	if ( this->behavior = &actPlayer )
```
addictgamer: *kaboom*
should be `==`

src/entity.cpp
```diff
@@ -7887,3 +7883,99 @@ Item** Entity::shouldMonsterEquipThisArmor(const Item& item) const
...
+		net_packet->address.host = net_clients[playerhit - 1].host;
+		net_packet->address.port = net_clients[playerhit - 1].port;
+		net_packet->len = 6;
+		sendPacketSafe(net_sock, -1, net_packet, playerhit - 1);
+	}
+}
+
+void Entity::removeLightField()
+{
+	if ( this->light != nullptr )
```
addictgamer: Too much of `this`!

src/monster_minotaur.cpp
```diff
@@ -834,6 +834,34 @@ void actMinotaurCeilingBuster(Entity* my)
...
+						else if (	entity->behavior == &actStalagCeiling	||
+									entity->behavior == &actStalagFloor		||
+									entity->behavior == &actStalagColumn 
+								)
+						{
+							// spawn several rock items
+							int c, i = rand() % 4;
+							for ( c = 0; c < i; ++c )
+							{
+								Entity* childEntity = spawnGib(my);
```
addictgamer: This should differentiate between the client and the server. You may recall the ceiling buster crash in multiplayer from a bit back.

src/entity.hpp
```diff
@@ -199,6 +199,18 @@ class Entity
 	//--PUBLIC AMBIENT PARTICLE EFFECT SKILLS--
 	Sint32& particleDuration;
 
+	//--PUBLIC DOOR SKILLS--
+	Sint32& doorDir;
```
addictgamer: We might start adding comments here as to what reference correlates to which skill. E.g. `Sint32& doorDir; //skill[0]`

src/actdoor.cpp
```diff
@@ -280,3 +281,29 @@ void actDoorFrame(Entity* my)
 		my->flags[PASSABLE] = true; // the actual frame should ALWAYS be passable
 	}
 }
+
+void Entity::doorHandleDamageMagic(int damage, Entity &magicProjectile, Entity *caster)
```
addictgamer: Perhaps damage should be calculated within this function. So a fire ball spell will deal more damage than magic missile, all calculated here internally.

Issue #115, opened 2017-08-14
[2.0.5][Multiplayer] Client disconnection is not handled gracefully


crkellen:
**ISSUE:**
When a Client in a multiplayer game disconnects, the game continues to believe they are still in the game. This causes them to load in when a new floor is loaded, before being removed after a brief delay (1 - 5 seconds). This causes multiple issues, and gets worse if more Clients disconnect.
***Some issues that can occur:***
- Clients still connected will see the wrong model for entities on the ground. They may attempt to pick up an apple, but receive a steel shield.
- Clients still connected may not be able to see chat with other Clients, in my case, I was the fourth Client, and the (I believe) second Client disconnected. The third Client could see my messages, I could not see his.
- If two Clients leave, the unexpected behavior gets worse. Many entities are now replaced with floating bronze maces standing upright above the ground, a bit above eye level. Some entities, such as spike traps and enemies, may be replaced with nothing, or other entities, such as items.

**EXPECTED RESULT:**
When a Client disconnects, a packet should be sent to handle the disconnect gracefully, alerting each player connected about the disconnect and removing them as necessary.

**STEPS TO REPRODUCE:**
1. Join a multiplayer game with 3 other players.
2. Start the game.
3. Have one Client disconnect.
4. Move down a floor, notice that the player still loads in briefly, before disappearing.
5. Explore, notice that on the Clients, certain items may appear to be other models.
6. Have another Client disconnect.
7. Move down another floor, notice that both players load in briefly, before disappearing.
8. Explore and notice the issues listed above.

**ADDITIONAL INFO:**
This seems like it may just be an issue of not alerting the Clients of a player's disconnection. This quickly spirals out of control, causing a lot of undefined behavior.

addictgamer:
Resolved in #205

PR #178, opened 2017-09-22
Fix Issue #115 - Handle Client Disconnect Gracefully

crkellen:
This is a fix for #115.
The issue was there was a `return` where there should have been a `continue`. This was causing disconnections to not get processed on any remaining Clients. It also prevented the Host from ever being notified that a Client disconnected.


addictgamer:
Merged in #205

PR #179, opened 2017-09-24
Move inline item functions to header

lheckemann:
Since they are inlined, their implementation needs to be included in
every source file that uses them, otherwise it will go missing entirely
and fail to link.


addictgamer:
Resolved in #169

Issue #163, opened 2017-08-18
[2.0.5][Multiplayer][Client] Clients do not have accurate positions of boulders


crkellen:
**ISSUE:**
Clients will see any boulder that is not perfectly aligned (on the grid system set for it) jittering in place rapidly and not in the correct position. Moving these boulders can often make the issue worse, causing them to be very out of sync.

**EXPECTED RESULT:**
Clients should see boulders are the host does, not jittering in place and causing issues.

**ADDITIONAL INFO:**
Reasons for this existing are related to #159.

addictgamer:
Resolved in #216

PR #181, opened 2017-09-26
Fix Issue #163 - Client Boulders Position Desync

crkellen:
This is a fix for #163.
The issue was that when a boulder was stopped, for any reason, nothing would set the velocity of the entity back to zero. This caused the host to constantly transmit a velocity, which would trigger the dead reckoning. The dead reckoning code would then attempt to move forward, but interpolate backwards, causing the constant vibrating of the boulder.

There are other entities that have this same issue. I believe the Devil being one of them. They will be fixed in a separate pull request.


addictgamer:
Merged in #216

Issue #182, opened 2017-09-26
[2.0.5][Multiplayer][Client] Client does not have accurate position of Devil


crkellen:
**ISSUE:**
When fighting the Devil on the 'hellboss' level, the Client will see the Devil surface from a pool of lava, then violently interpolate to a different location.

**EXPECTED RESULT:**
The position of the Devil should be accurate, and should not attempt to interpolate to a invalid position.

**STEPS TO REPRODUCE:**
1. Edit levels.txt to start at 'hellboss'
2. Start a multiplayer game
3. Watch on the Client as the Devil will violently attempt to interpolate to (the player starting position?) another position when surfacing in a lava pool.

**ADDITIONAL INFO:**
This is the same as #163. The cause is the velocity of the entity not being reset to 0.

addictgamer:
Resolved in #217

PR #183, opened 2017-09-26
Fix Issue #182 - Client Devil Position Desync

crkellen:
This is a fix for #182.
The issue was that the Devil's velocity needs to be updated in order to calculate the yaw to face the Player. Leaving this velocity as a non-zero value will cause dead reckoning on the Client to move the Devil. The velocity never needs to be a value other than 0 after calculating the yaw, so it can be safely reset to zero after the calculation.


addictgamer:
Merged in #217

PR #184, opened 2017-09-26
Fix Issue #141 - Multiplayer Magic Missile Cost

crkellen:
This is a fix for #141.
The issue was in Multiplayer there was not an additional check to see if the Caster had 100 proficiency in Spellcasting while casting Magic Missile. There needed to be an additional one here as 'getCostOfSpell()' does not have access to the Player proficiency.


addictgamer:
`//TODO: Fix issue #141: Spellcasting capstone doesn't work in multiplayer.`

...apparently I already did such. Wow, I completely forgot.

crkellen:
You added that TODO shortly after I pointed out that it didn't work. I checked the commit history last week.
null

addictgamer:
> You added that TODO shortly after I pointed out that it didn't work.

Ya, for some reason I didn't remember doing that until I reviewed the commit history again last night.

addictgamer:
Good.

PR #185, opened 2017-09-27
Model cache

lheckemann:
PRing for attention/comments/utility to people even though it's not release-ready. This is based on (and has a small dependency on) #86, but the key addition is 41250d4. This change comes with a number of caveats, documented extensively in the commit message, but the benefit is that it makes Barony load **7 times faster** once the cache has been generated on the first launch.


WALLOFJUSTICE:
This is great! I just cherry-picked the Cache models commit 41250d4 and edited the openUserFile with the old openDataFile to get it running.

I've added locally /usemodelcache as a flag in default.txt to decide whether to use the cache, and /disablecache in game to clear the flag for the next run.

All that's missing is a rebuild section in game (something like /regenmodelcache then I could be content with it. Having an optional flag in default.txt (false by default - user must enter command or edit the file manually) makes it OK to pull IMO.

WALLOFJUSTICE:
My work is on the models-cache branch, commits 335718e3c126c284fdfcd92a9b8be98d229115e6 and 74227f0d8575e52e2f039f657a5e2d04b064c13c

lheckemann:
What may also make sense is to ship individual model files in the cached format consisting of triangles, or append the triangles to the voxel model files. I'll draft something these next few days.

addictgamer:
> rather than storing them all in one big cache blob file. 

Any particular reason why you'd prefer that route? I'm fine with one blob file if it means less disk I/O throughout the process.

lheckemann:
The main issue with the cache blob file is that it doesn't keep the polygon data in a relationship with its voxel "source" data, which means we'd have to mess around with hashes and stuff, or suffer the longer original load times when modifying the models (and there's no real technical reason for that).

As for disk access, it's not a noticeable factor in load times for Barony. Keep in mind that all the file loading is done *before* the progress indicator for models is displayed, so what we're addressing here is very much a CPU/memory-bottlenecked process.

addictgamer:
>  Keep in mind that all the file loading is done before the progress indicator for models is displayed

Ah, nevermind that then.

lheckemann:
Integrated in ec76d925cacb688a5cc2ee719a7a862b037309f1

PR #186, opened 2017-10-08
* Fix for resuming savegames in multiplayer, seed was being reset to 0 on map reload.

WALLOFJUSTICE:
* Fix for resuming savegames in multiplayer, seed was being reset to 0 on map reload.

* Version number bumped to 2.0.6


addictgamer:
Looks like github didn't detect the merge, but this is now in the 2.0.6-branch branch.

PR #200, opened 2017-10-12
[MASTER] Top border for stat bars

crkellen:
See #113.

### DIFF
--- a/src/interface/drawstatus.cpp
+++ b/src/interface/drawstatus.cpp
@@ -480,7 +480,7 @@ void drawStatus()
 	{
 		pos.x = 80;
 		pos.w = 33;
-		pos.h = 129 * (static_cast<double>(stats[clientnum]->HP / stats[clientnum]->MAXHP));
+		pos.h = 129 * (static_cast<double>(stats[clientnum]->HP) / stats[clientnum]->MAXHP);
 		pos.y = yres - 15 - pos.h;
 
 		if ( stats[clientnum]->EFFECTS[EFF_POISONED] )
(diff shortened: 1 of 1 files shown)


src/interface/drawstatus.cpp
```diff
@@ -430,35 +430,52 @@ void drawStatus()
...
 	pos.y = yres - 168;
 	drawTooltip(&pos);
+
+	// Display "HP" at top of Health bar
 	ttfPrintText(ttf12, pos.x + 8, pos.y + 6, language[306]);
 
+	// Display border between actual Health bar and "HP"
+	pos.x = 76;
+	pos.w = 38;
+	pos.h = 1;
```
WALLOFJUSTICE: Change height from 1 to 0 - Makes the border 2px high instead of 3px. Matches the 2px border for the rest of the bar.

src/interface/drawstatus.cpp
```diff
@@ -480,18 +497,33 @@ void drawStatus()
...
 	pos.y = yres - 168;
 	drawTooltip(&pos);
+
+	// Display "MP" at the top of Magic bar
 	ttfPrintText(ttf12, pos.x + 8, pos.y + 6, language[307]);
+
+	// Display border between actual Magic bar and "MP"
+	pos.x = 12;
+	pos.w = 39;
+	pos.h = 1;
```
WALLOFJUSTICE: Change height from 1 to 0 - Makes the border 2px high instead of 3px. Matches the 2px border for the rest of the bar.

crkellen:
I will handle this today.
null

crkellen:
In addition to correcting the box heights, I realized that the HP bars are actually not centered in the X or Y axis. I corrected this, and also added a small amount of green value to the MP bar to make it significantly less jarring to look at. Finally, I increased the faint background bars' color from 16 -> 48

WALLOFJUSTICE:
Fine with me!

addictgamer:
Thanks

PR #201, opened 2017-10-12
[MASTER] Fix Issue #82 - Pushed boulder can collide with pusher 

crkellen:
See #159.

### DIFF
--- a/src/actboulder.cpp
+++ b/src/actboulder.cpp
@@ -28,33 +28,60 @@
 #define BOULDER_DESTX my->skill[6]
 #define BOULDER_DESTY my->skill[7]
 
-/*-------------------------------------------------------------------------------
-
-	boulderCheckAgainstEntity
-
-	causes the boulder given in my to crush the object given in entity
-	or vice versa
-
--------------------------------------------------------------------------------*/
-
-int boulderCheckAgainstEntity(Entity* my, Entity* entity)
+/* actboulder.cpp
+ * @param my - A pointer to the Boulder which is checking for collisions
+ * @param entity - A pointer to the current Entity being checked against @my
+ * @param bWasBoulderPushed - True if the Boulder was pushed before calling this function
+ * @returns true - If the Boulder was destroyed as a result of a collision
+ * @returns false - Default case
+ * Checks for collisions against every Entity on the map
+ * Damages Players or Monsters. Can be destroyed if the Entity survives
+ * Destroys Doors and stops movement in every other case
+ */
+bool BoulderCheckAgainstEntity(Entity* const my, Entity* const entity, const bool bWasBoulderPushed)
 {
-	if (!my || !entity)
+	if ( my == nullptr || entity == nullptr )
 	{
-		return 0;
+		printlog("ERROR: BoulderCheckAgainstEntity() - A parameter is null. my = %d, entity = %d.", (my == nullptr), (entity == nullptr));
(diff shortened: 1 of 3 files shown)


src/actboulder.cpp
```diff
@@ -235,7 +296,7 @@ void actBoulder(Entity* my)
 					{
 						continue;
 					}
-					if ( boulderCheckAgainstEntity(my, entity) )
+					if ( BoulderCheckAgainstEntity(my, entity) )
```
WALLOFJUSTICE: Compile error in VS, needs 3 arguments:
Error	C2660	'BoulderCheckAgainstEntity': function does not take 2 arguments		

bool BoulderCheckAgainstEntity(Entity* const my, Entity* const entity, const bool bWasBoulderPushed)

src/actboulder.cpp
```diff
@@ -338,7 +399,7 @@ void actBoulder(Entity* my)
...
```
WALLOFJUSTICE: Compile error in VS, needs 3 arguments:
Error	C2660	'BoulderCheckAgainstEntity': function does not take 2 arguments	

bool BoulderCheckAgainstEntity(Entity* const my, Entity* const entity, const bool bWasBoulderPushed)

src/actboulder.cpp
```diff
@@ -507,7 +568,7 @@ void actBoulder(Entity* my)
 						{
 							continue;
 						}
-						if ( boulderCheckAgainstEntity(my, entity) )
+						if ( BoulderCheckAgainstEntity(my, entity) )
```
WALLOFJUSTICE: Compile error in VS, needs 3 arguments:
Error	C2660	'BoulderCheckAgainstEntity': function does not take 2 arguments	

bool BoulderCheckAgainstEntity(Entity* const my, Entity* const entity, const bool bWasBoulderPushed)

crkellen:
My bad. I rushed the new PR and messed it up. Will fix today.
null

src/collision.cpp
```diff
@@ -258,6 +258,87 @@ bool entityInsideEntity(Entity* entity1, Entity* entity2)
...
+					{
+						if ( (entity1->y + entity1->sizey) > (entity2->y - entity2->sizey) )    // Check everything South (+Y) Is the Entity not too far to the right of the boulder?
+						{
+							return true;
+						}
+					}
+				}
+			}
+			return false;
+			break;
```
addictgamer: A break right after a return is redundant

WALLOFJUSTICE:
Managed to get myself stuck on a boulder.., Pushing the boulder with myself against the wall was fine, but moving forward + left into empty space while mashing push made the boulder snap back to it's spot while my player managed to partially occupy the tile
image

crkellen:
I'll investigate this edge case.
null

crkellen:
Yep! Must have forgotten to remove it. VS doesn't flag the "any code past a return will not be executed". Will fix today.
null

crkellen:
@WALLOFJUSTICE I was able to partially replicate your issue twice. I do not know of a way to reliably reproduce it. In both of my cases, I was prevented from moving left, however I could simply move backwards or to move right in order to break free.

A better solution would be to calculate the path of the boulder ahead of time, which I thought of, but that could cause an excessive amount of issues with things moving out of the way or into the way of a rolling boulder.

PR #202, opened 2017-10-12
[MASTER] Fix Issue #100 - Client cannot see themselves "On Fire" 

crkellen:
See #122.


addictgamer:
Hmm. Actually, it seems to me as if the cure ailment spell isn't resolving fire on the client's side properly.
It'll say:
```
It burns! It burns!
Wow! You feel restored.
It burns! It burns!
It burns! It burns!
...
asdf burns to a crisp.
You die...
```

Other times, it seems to work fine.

crkellen:
I had forgotten to also investigate the issue with the potion as well. I'll do that today, since I get off early.
null

PR #207, opened 2017-10-12
[MASTER] Fix Issue #120 - Cure Ailment does not remove Fire on Clients

crkellen:
See #132.

### DIFF
--- a/src/item_usage_funcs.cpp
+++ b/src/item_usage_funcs.cpp
@@ -379,10 +379,10 @@ void item_PotionCureAilment(Item* item, Entity* entity)
 		}
 		return;
 	}
-	if ( players[clientnum]->entity->flags[BURNING] )
+	if ( players[player]->entity->flags[BURNING] )
 	{
-		players[clientnum]->entity->flags[BURNING] = false;
-		serverUpdateEntityFlag(players[clientnum]->entity, BURNING);
+		players[player]->entity->flags[BURNING] = false;
+		serverUpdateEntityFlag(players[player]->entity, BURNING);
 	}
 	if ( multiplayer == CLIENT )
 	{
--- a/src/magic/castSpell.cpp
+++ b/src/magic/castSpell.cpp
@@ -645,7 +645,7 @@ Entity* castSpell(Uint32 caster_uid, spell_t* spell, bool using_magicstaff, bool
 					if ( players[i]->entity->flags[BURNING] )
 					{
 						players[i]->entity->flags[BURNING] = false;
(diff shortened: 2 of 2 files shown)


crkellen:
I do believe I fixed cure ailment potion as well. That fix may be hiding somewhere. I will add it to this PR likely later today.
null

crkellen:
Apologies that this took so long. I had forgotten about it for a little bit there. All fixed!

addictgamer:
Potion of Cure Ailment works now, however:

> On use of 'cure ailment' spell, the flame particles are still present on the server player model and client player.

Potion of Cure Ailment stops the burning and makes the flame particles disappear on both the client's end and the server's end.
Spell of Cure Ailment stops the burning but only makes the flame particles disappear on the server's end.

WALLOFJUSTICE:
Confirming that the spell of cure ailment cast by the client will only remove flames on the host.

If spell is cast by host, all players will have the flame effect removed from their views. Just need to get this 1 test case to solve.

crkellen:
I may have incorrectly copy-pasted. I'll review today.
null

crkellen:
The flames will Continue to be present as they will be fixed in the other pull request.
null

addictgamer:
Ah. Is that in an existing PR or will you be making a new one?

crkellen:
Already existing. If not then I never created it. I have a branch for it  sure. Heading home now, so I can handle this.
null

addictgamer:
Right. I just tested this branch again based off of the latest master. The issue still occurs though for the cure ailment spell.

The cure ailments potion does makes the flames go away, the spell doesn't seem to be doing such.

@WALLOFJUSTICE can you confirm if the issue still occurs for you or if it's fixed?

crkellen:
I can check on this tomorrow.
null

crkellen:
OKay, I can totally confirm that I simply messed up somewhere down the line and did this:
Current:
```
if ( players[i]->entity->flags[BURNING] )
{
    players[i]->entity->flags[BURNING] = false;
    serverUpdateEntityFlag(players[clientnum]->entity, BURNING);
}
```

When it should be:
```
if ( players[i]->entity->flags[BURNING] )
{
    players[i]->entity->flags[BURNING] = false;
    serverUpdateEntityFlag(players[i]->entity, BURNING);
}
```

I believe I redid the PR or messed up somewhere, because I do remember testing it and having it work just fine, but clearly I somehow left it out of the actual PR. I'm going to test it real quick just to be 100% then i'll push the updated commit.

addictgamer:
Confirm fixed.

PR #215, opened 2017-10-12
[MASTER] Fix Issue #156 - Disable Steam Achievements if Cheats are enabled

crkellen:
See #168.

### DIFF
--- a/src/steam.cpp
+++ b/src/steam.cpp
@@ -613,6 +613,12 @@ void steamAchievement(const char* achName)
 	return;
 #else
 
+	if ( !(svFlags & SV_FLAG_CHEATS) )
+	{
+		// Cheats are enabled, therefore you cannot earn Steam Achievements
+		return;
+	}
+
 	if ( !achievementUnlocked(achName) )
 	{
 		//messagePlayer(clientnum, "You've unlocked an achievement!\n [%s]",c_SteamUserStats_GetAchievementDisplayAttribute(achName,"name"));
@@ -642,6 +648,12 @@ void steamAchievementClient(int player, const char* achName)
 {
 	if ( multiplayer == CLIENT )
 	{
+		return;
+	}
+
+	if ( !(svFlags & SV_FLAG_CHEATS) )
+	{
+		// Cheats are enabled, therefore you cannot earn Steam Achievements. This check prevents needless packet sending
 		return;
 	}
 
--- a/src/steam.cpp
+++ b/src/steam.cpp
@@ -613,7 +613,7 @@ void steamAchievement(const char* achName)
 	return;
 #else
 
-	if ( !(svFlags & SV_FLAG_CHEATS) )
+	if ( svFlags & SV_FLAG_CHEATS )
 	{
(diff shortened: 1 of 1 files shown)


src/steam.cpp
```diff
@@ -613,6 +613,12 @@ void steamAchievement(const char* achName)
 	return;
 #else
 
+	if ( !(svFlags & SV_FLAG_CHEATS) )
```
WALLOFJUSTICE: The if statements need the ! removed. This implements the opposite effect, you only get Steam Achievements when cheats are on. 

For reference, when cheats are enabled, svFlags = 31 and SV_FLAG_CHEATS is constant 1. `!svFlags & SV_FLAG_CHEATS` equals 0, which skips the return label and gives the player the achievement. When cheats disabled, then svFlags = 30, `!svFlags & SV_FLAG_CHEATS` equals 1, which returns. 

`if ( !(svFlags & SV_FLAG_CHEATS) )
{
//code
}` should be 
`if ( (svFlags & SV_FLAG_CHEATS) )
{
//code
}`

src/steam.cpp
```diff
@@ -645,6 +651,12 @@ void steamAchievementClient(int player, const char* achName)
 		return;
 	}
 
+	if ( !(svFlags & SV_FLAG_CHEATS) )
```
WALLOFJUSTICE: if ( !(svFlags & SV_FLAG_CHEATS) ) { //code } should be
if ( (svFlags & SV_FLAG_CHEATS) ) { //code }

src/steam.cpp
WALLOFJUSTICE: I'll note the bitwise operations are very hard to visually determine what's going on. I'll admit it looked correct before I convinced myself otherwise by doing a playtest.

crkellen:
This is partially due to my inexperience with bitwise operations, and more so because I merely copied it from the /commands section of the code and forgot to reverse it.
I will fix it today.
null

crkellen:
I will investigate adding those extra enhancements when I can.

WALLOFJUSTICE:
Alrighty lets get this one tidied up. Bring it up to barony-next for a start.

Let's stuff a ifdef STEAMWORKS around a tooltip like the traps note. "NOTE: Enabling cheats will disable Steam achievements!"

Then above our v3.0.0 text in the bottom right, stuff a ifdef STEAMWORKS and print out the string "Cheats turned on, Steam achievements are disabled!"

image

crkellen:
I can handle this Friday.
null

crkellen:
Can I make a slight modification to your request?
The current implementation uses the lang file to display the text boxes. So added the line `NOTE: Enabling cheats will disable Steam achievements!` to the lang file doesn't require any modification for the code at all. The modification to your request being that there is no IFDEF, and even if the Player is using non-steam, they will still see the message.

However, I can still add the IFDEF to the text above the version string.
Otherwise, if you want, I can find a way to ifdef it in there.

crkellen:
This is the MASTER branch PR, please see #168.

PR #217, opened 2017-10-12
[MASTER] Fix Issue #182 - Client Devil Position Desync 

crkellen:
See #183.


crkellen:
Never noticed. I can see if I can see that on my end later.
null

crkellen:
This issue is also present on master.

WALLOFJUSTICE:
Since the sliding issue was present before and after this PR, then this change is approved.

PR #224, opened 2017-10-29
Player class weighted stat growths

WALLOFJUSTICE:


src/entity.cpp
```diff
@@ -8650,3 +8620,34 @@ char* Entity::getMonsterLangEntry()
 	}
 	return nullptr;
 }
+
+void playerStatIncrease(int playerClass, int chosenStats[3])
+{
+	std::mt19937 seed(rand()); // seed of distribution.
```
addictgamer: So this function is getting reseeded every time the player levels up?

Issue #219, opened 2017-10-29
Poison kills Buddha


addictgamer:
Poison can kill the player in buddha mode. Shouldn't.

addictgamer:
Buddha should now never die from any source ever:
fdc611e161baf77d6da5b78829e163eb3c5f683a

Issue #220, opened 2017-10-29
Hungry Buddha Dies


addictgamer:
Players can starve to death in buddha mode

addictgamer:
Fixed in #220

PR #225, opened 2017-10-31
[MASTER] Fix Issue #219 - Poison Kills Buddha

crkellen:
This is a fix for #219.
The issue is that `modHP()` does not do any checks for `godmode` or `buddhamode`. Adding a check there would be bad as `buddhamode` normally leaves you with 1 HP at the least. Changing modHP will affect every source of HP modifications, and that is why the check is made in `handleEffects()`.

### DIFF
--- a/src/entity.cpp
+++ b/src/entity.cpp
@@ -1788,17 +1788,28 @@ void Entity::handleEffects(Stat* myStats)
 		{
 			this->char_poison = 0;
 			int poisonhurt = std::max(1 + rand() % 4 - myStats->CON, 3);
-			this->modHP(-poisonhurt);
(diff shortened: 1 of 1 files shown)


addictgamer:
Same issue as with #220, player will only take poison damage if buddha mode is enabled.
Can confirm the poisoned buddha won't die, at least.

crkellen:
Once again, will solve tomorrow. My mistake for not doing extensive testing.
null

crkellen:
This has been solved. Please review when you have the time @addictgamer.

crkellen:
Rebased to barony-next in #283.

PR #226, opened 2017-11-01
[MASTER] Fix Issues #220 - Hungry Buddha Dies

crkellen:
This is a fix for #220.
The issue is that `modHP()` does not do any checks for `godmode` or `buddhamode`. Adding a check there would be bad as `buddhamode` normally leaves you with 1 HP at the least. Changing modHP will affect every source of HP modifications, and that is why the check is made in `handleEffects()`.

Additionally, I believe it would be beneficial for taking damage from hunger to cause a screen shake and a damage noise. I can add that to this PR, or to a separate one.

### DIFF
--- a/src/entity.cpp
+++ b/src/entity.cpp
@@ -1582,6 +1582,17 @@ void Entity::handleEffects(Stat* myStats)
 							this->setHP(1);
 						}
 					}
+					else
+					{
+						this->modHP(-4);
+
+						if ( myStats->HP > 0 )
+						{
+							messagePlayer(player, language[633]);
+						}
+
+						this->setObituary(language[1530]);
+					}
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


addictgamer:
The code doesn't look right. The player will only take hunger damage if buddha mode is enabled.
Can confirm that the hungry buddha doesn't die, at least.

crkellen:
Oh wow. That's pretty funny. My bad, I'll solve this tomorrow.
null

crkellen:
This has been solved. Please review when you can @addictgamer.

WALLOFJUSTICE:
Again, up to barony-next for the credit of the PR, then I'll test.

crkellen:
Rebased to barony-next in #284.

PR #227, opened 2017-11-02
[MASTER] Enhancement - Hunger damage plays sound and shakes screen

crkellen:
This is an enhancement for when a Player takes damage from starvation.
The issue is that there is no screen shake or damage sound played when you take damage from starvation. This is an issue because every other source of damage causes the screen to shake and a sound clip to play.

This issue is fixed by adding in a screen shake and playing `Damage.ogg` using the same parameters as taking damage from poison.


addictgamer:
Starving is much better now.

Issue #75, opened 2017-05-26
Humans shouldn't equip stuff when sleeping


addictgamer:
Throw equipment at a sleeping human and he'll equip himself in his sleep. Wish I could do that.

addictgamer:
Fixed in #285

PR #228, opened 2017-11-02
[MASTER] Fix Issue #75 - Entities can equip Items while Asleep

crkellen:
This is a fix for #75.
The issue is that there are no checks to prevent Entities from equipping Items. Additional checks could be added such as Paralysis.


WALLOFJUSTICE:
Let's push this one up to barony-next as well, conflicts abound if we go master. Easier to give you credit for it

crkellen:
Understood.
null

crkellen:
Rebased to barony-next in #285.

Issue #73, opened 2017-05-26
Shopkeepers sometimes don't respond to attacks


addictgamer:
Might be due to flooding his shop with followers? Related to minotaur bug?

crkellen:
It seems the issue lies with distance. I haven't fully tracked it down yet, but distance seems to be a major part of it. I'll continue to track this one down and hopefully have it resolved soon.

WALLOFJUSTICE:
Addressed in 457cbd69dd61c545d998b08af73a6ecfb4428905 for melee attacks on all friendlies

PR #229, opened 2017-11-03
[MASTER] Fix Issue #73 - Shopkeeper Hostility does not consistently trigger

crkellen:
This is a fix for #73.
The issue is that the code that triggers Shopkeeper's hostility towards Players is only performed in the "Charge" state. If a Player attacks a stationary Shopkeeper from up close, the Shopkeeper will move towards the Player, but will not trigger any aggression. However, if a Player is slightly further away, the Shopkeeper will move towards the Player, allowing the code to trigger, setting Player's as `swornenemies[SHOPKEEPER][HUMAN]`.

This is fixed by extracting the code out of the state machine. This allows for hostility to be triggered during any state. This does not fix the exact same issue for Humans (potential Followers and the like), which will be addressed in a separate PR.

### DIFF
--- a/src/actmonster.cpp
+++ b/src/actmonster.cpp
@@ -1683,6 +1683,29 @@ void actMonster(Entity* my)
 			}
 		}
 
+		// Check if Shopkeepers are now hostile against Players
+		if ( myStats->type == SHOPKEEPER )
+		{
+			Entity* targetedEntity = uidToEntity(MONSTER_TARGET);
+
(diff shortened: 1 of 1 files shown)


WALLOFJUSTICE:
Looks good, can we get this as a barony-next PR since we won't be jumping back to master soon?

crkellen:
Yes. I can handle it Friday.
null

crkellen:
Rebased to barony-next in #286.

Issue #230, opened 2017-11-03
[2.0.7][AI] Humans do not always turn hostile when attacked


crkellen:
This is related to #73, however the implementation is different.
An explanation of the issue is given in #229 and is quoted below.

> The issue is that the code that triggers Shopkeeper's hostility towards Players is only performed in the "Charge" state. If a Player attacks a stationary Shopkeeper from up close, the Shopkeeper will move towards the Player, but will not trigger any aggression. However, if a Player is slightly further away, the Shopkeeper will move towards the Player, allowing the code to trigger, setting Player's as swornenemies[SHOPKEEPER][HUMAN].
> 
> This is fixed by extracting the code out of the state machine. This allows for hostility to be triggered during any state. This does not fix the exact same issue for Humans (potential Followers and the like), which will be addressed in a separate PR.

Humans do not change `swornenemies[][]` as that would change it so that every Human is hostile to every Player at all times. Extracting the code will be a bit more difficult.

PR #231, opened 2017-11-08
Fix Issue #230 - Humans' Hostility does not consistently trigger 

crkellen:
This is a fix for #230.
The issue is the same as #229, but not as easily solved as Humans can become Followers, and don't change the `swornenemies[][]` array. A similar solution was taken to solve this issue. However, there is still one issue left:

Due to the "Call Allies to Aid" functionality, Followers of a Player that attacks a Human will also become aggressive against that Player. The changes in this PR however do successfully remove the Player from the 'FOLLOWING' list.

### DIFF
--- a/VS.2015/Barony/Barony.vcxproj.filters
+++ b/VS.2015/Barony/Barony.vcxproj.filters
@@ -249,65 +249,8 @@
     <ClCompile Include="..\..\src\player.cpp">
       <Filter>Source Files</Filter>
     </ClCompile>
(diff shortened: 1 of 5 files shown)


WALLOFJUSTICE:
Could you do a quick pull from the barony-next branch to try get this up to date? Hopefully it'll sort out any conflicts.

crkellen:
Yes, I can update this and see what I can do.
null

Issue #84, opened 2017-06-08
Animated tiles with numbering 0-9 at the filename end count as swimming tiles


WALLOFJUSTICE:
Makes sense for the water/lava frames, but you swim in the glowing hell and citadel pillars if they are placed as ground tiles. Not an issue in the main maps as these animated blocks are for pillars, but affects mapmakers.

TODO: Merge the swimming checks into a common function, change the way we handle swimming tiles to only affect water/lava. Info is determined in init_game.cpp during the tiles.txt read. Stores the swimming result in animatedTiles[] which is referenced for subsequent swimming checks.

crkellen:
This can likely be fixed with #144, as the addition of an Entity flag would allow for better handling of swimming checks. Commenting to implement later.

addictgamer:
Fixed in #235

PR #235, opened 2017-11-13
Fix animated tiles being swimmable.

WALLOFJUSTICE:
+ swimmingtiles
+ vertical spell trap
+ ceiling tile model in editor


addictgamer:
Fixes #84?

WALLOFJUSTICE:
Yes

WALLOFJUSTICE:
Should also note anything with lava in the name is still lava, affects 1 or 2 animated tiles so they can be renamed if necessary

Issue #221, opened 2017-10-29
Monsters using slingshots give player wrong hit message


addictgamer:
it says  you are "hit by an arrow!"
Should say "You are hit by a stone!"

addictgamer:
Fixed in #238

PR #237, opened 2017-11-16
Fix Issue #221 - Hit by Slingshot gives wrong message

crkellen:
This is a fix for #221.
The issue is that there is no separate lang entry for being hit by a stone. Additionally, in `actArrow()` there is no separate usage, since there is no entry. I added an entry, and a usage that can be scaled to accommodate more types of projectiles.

### DIFF
--- a/VS.2015/Barony/Barony.vcxproj.filters
+++ b/VS.2015/Barony/Barony.vcxproj.filters
@@ -249,65 +249,8 @@
     <ClCompile Include="..\..\src\player.cpp">
       <Filter>Source Files</Filter>
     </ClCompile>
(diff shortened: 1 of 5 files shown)


WALLOFJUSTICE:
I've addressed this issue locally already, apologies for you spending your time on this

crkellen:
You know, I had a feeling you did, but didn't remember seeing any commit addressing it. Honestly, I should have did a double take to make sure. It's not like it took very long anyway.
null

addictgamer:
#238 has been merged in which, as discussed above, fixes this issue.

PR #240, opened 2017-11-24
Move sprites with selection tool in editor, new "pencil" button in editor

WALLOFJUSTICE:
+ Add new map flag for maps to have alternate generation.


addictgamer:
Tiles after 213 have no tooltips?

addictgamer:
If I select an entity and then alt+arrow the selection around and stop when I've over some other entities, if I alt+arrow again then, it won't pick up the new entities until I either resize it or ctrl+arrow around a bit.

PR #246, opened 2017-11-26
Fix which it seems just a typo for the OpenAL build.

devnexen:


crkellen:
Note, this fix seems to be required to build barony-next now.

addictgamer:
Good catch.

Issue #244, opened 2017-11-26
Revived client doesn't start with half of max MP


addictgamer:
When entering the next level, revived clients should start with half of their max MP.
Works fine for hosts and in singleplayer.

addictgamer:
Fixed in #248

PR #248, opened 2017-11-27
Fix Issue #244 - Client does not respawn with half MP

crkellen:
This is a fix for #244.
The issue is that there was just a missing line of code to match the Host's code of setting the Client's MP to be 1/2 that of their max.


WALLOFJUSTICE:
Approved. My party stats feature made this less of an issue as it continually refreshes client's MP values every 3 seconds, but this will correctly set as soon as you spawned in.

Issue #249, opened 2017-11-27
[2.0.7] Fireball can set friendlies on fire with "Friendly Fire" off


crkellen:
**ISSUE:**
When the "Friendly Fire" flag is set to Off, meaning Players cannot harm their allies, hitting a friendly with a Fireball spell or staff will still set the friendly on fire, causing them to take damage.

**EXPECTED RESULT:**
If "Friendly Fire" is Off, there should be no method of harming a friendly through first-party means. All other spells effects do nothing if the friendly is hit, Fireball should be the same.

**STEPS TO REPRODUCE:**
1. Start a multiplayer game
2. Turn Off "Friendly Fire"
3. Cast fireball or shoot a staff of fireball at another Player, they will be set on fire, and will take damage.

**ADDITIONAL INFO:**
I recommend this for a hotfix, if at all possible. Perhaps with a rework on Fire to make it non-random, but have some sort of maximum. See #121.

addictgamer:
Fixed in #252

PR #250, opened 2017-11-28
Fix Issue #249 - Fireball can set friendlies on fire with F.F. off

crkellen:
This is a fix for #249.
The issue is that the Entity is set on fire before checking to see what Entity is. This prevents the check for friendly fire from happening before the Entity. I placed the code that sets the Entity on fire at the end, which will prevent the spell from setting friendly Players or Monsters (Humans/Shopkeepers) on fire.


crkellen:
Closing this to be replaced by #252.

Issue #241, opened 2017-11-26
Minimap draws over tooltips


addictgamer:
Tooltips should draw over minimap

crkellen:
Do you want the Status Bar to draw over the minimap as well? Or just the tooltips?
The issue is that the function `drawStatus()` is being called before `drawMinimap()`.

addictgamer:
Resolved in #253

PR #253, opened 2017-11-29
Fix Issue #241 - Minimap draws over Status Bar

crkellen:
This is a fix for #241.
The issue is that the Status Bar, which also draws Tooltips, executes before `drawMinimap()`. This simple rearrangement should be sufficient. Due to placing the `drawStatus()` call after a few other code blocks, there could be other side effects. This can be solved by rearranging a bit more.


WALLOFJUSTICE:
It's legit, approved 
image

Issue #242, opened 2017-11-26
Breaking a chest while it's opened should close it


addictgamer:
Instead, this happens:
![break-chest-while-opened](https://user-images.githubusercontent.com/346089/33237097-55c90f0c-d21e-11e7-9b9b-96f8e1ec484f.png)

addictgamer:
Fixed in #254

PR #254, opened 2017-11-30
Fix Issue #242 - Chests should close before being broken

crkellen:
This is a fix for #242.
The issue is there was no call to `this->closeChest()` anywhere within the death code. Adding one solves the issues for both Client and Server.

However, for the one in the chest's inventory, they receive the message "You slam the chest shut". An alternate message could be implemented fairly easily.


addictgamer:
Fixes the issue, but the player still receives the message `you slam the chest shut!`

crkellen:
Sorry. I didn't add a new message, I had considered it, but at the time neglected to ask you what the message should say.
null

crkellen:
Was going to handle last Sunday, but was unable to load the build without assets. Will handle Friday.
null

crkellen:
This should be good to go now. I simply used entry `671 The chest is smashed into pieces!#`.

PR #255, opened 2017-12-01
Misc additions/balance

WALLOFJUSTICE:


src/actchest.cpp
```diff
@@ -1100,3 +1141,24 @@ void Entity::lockChest()
 {
 	chestLocked = 1;
 }
+
+void Entity::chestHandleDamageMagic(int damage, Entity &magicProjectile, Entity *caster)
+{
+	chestHealth -= damage; //Decrease door health.
```
addictgamer: "door" :)

src/draw.cpp
```diff
@@ -1829,6 +1829,18 @@ SDL_Rect ttfPrintTextColor( TTF_Font* font, int x, int y, Uint32 color, bool out
 		}
 	}
 
+	Uint32 cacheLimit = 8096;
```
addictgamer: Should be moved to a header. Maybe exposed as a config option.

PR #256, opened 2017-12-02
Cmake clean up

devnexen:
I started to work on packaging barony for *BSD systems thus the need to be able to install them properly with the help of the handy GNUInstallDir plugin, possibility to rename editor binary to avoid possible conflicts. I did not apply purposely for Darwin machines as I  was not sure @addictgamer would like (probably on these, only GOG and Steam are planned ... not home-brew/MacPorts ?).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.6)
+cmake_minimum_required(VERSION 2.8)
 project(barony)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_SOURCE_DIR}/cmake/Modules)
@@ -36,6 +36,9 @@ endif()
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/Modules/")
 
 set(CMAKE_OSX_ARCHITECTURES x86_64)
+
+set(EDITOR_EXE_NAME "editor" CACHE STRING "Editor executable name")
+option(DATA_DIR "Points to install data directory" FALSE)
 
 if (STEAMWORKS_ENABLED)
 message("------------------------")
@@ -353,17 +356,26 @@ if (OPENAL)
   endif()
 endif()
 
+set(BASE_DATA_DIR "./" CACHE INTERNAL "Base data dir")
 
-if (NOT APPLE)
-#install(TARGETS barony
-#  RUNTIME DESTINATION bin
-#  COMPONENT Runtime
-#)
+if (NOT APPLE AND UNIX)
+include(GNUInstallDirs)
+if (DATA_DIR)
+	set(BASE_DATA_DIR "${CMAKE_INSTALL_FULL_DATAROOTDIR}/barony")
+endif()
+message(STATUS "Base data directory ${BASE_DATA_DIR}")
+
+install(TARGETS barony
+	RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+  COMPONENT Runtime
+)
 else()
 install(TARGETS barony
   RUNTIME DESTINATION bin COMPONENT Runtime BUNDLE DESTINATION Resources
 )
 endif()
+
+add_definitions(-DBASE_DATA_DIR=\"${BASE_DATA_DIR}\")
 
 if (APPLE)
   #add_executable(editor OSX/SDLmain.m ${EDITOR_SOURCES})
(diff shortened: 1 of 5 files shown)


devnexen:
Is there a little chance it makes its way to 3.0.0 :-) ? (after the beta "craziness" of course)

addictgamer:
We'll start reviewing PRs after everything settles down :)

CMakeLists.txt
```diff
@@ -353,91 +356,112 @@ if (OPENAL)
...
 
+set(BASE_DATA_DIR "./" CACHE INTERNAL "Base data dir")
 
-if (NOT APPLE)
-#install(TARGETS barony
-#  RUNTIME DESTINATION bin
-#  COMPONENT Runtime
-#)
+if (NOT APPLE AND UNIX)
+include(GNUInstallDirs)
```
lheckemann: This should probably be indented for clarity's sake

CMakeLists.txt
```diff
@@ -353,91 +356,112 @@ if (OPENAL)
   endif()
 endif()
 
+set(BASE_DATA_DIR "./" CACHE INTERNAL "Base data dir")
```
lheckemann: Would be nice to have this as an option of some sort too. Not sure there's a nice way to provide an interface for this using cmake though…

CMakeLists.txt
```diff
@@ -353,91 +356,112 @@ if (OPENAL)
...
+add_definitions(-DBASE_DATA_DIR=\"${BASE_DATA_DIR}\")
+
 if (APPLE)
   #add_executable(editor OSX/SDLmain.m ${EDITOR_SOURCES})
   #set(MACOSX_BUNDLE_ICON_FILE "${PROJECT_SOURCE_DIR}/editor.icns")
   #add_executable(editor MACOSX_BUNDLE OSX/SDLmain.m ${EDITOR_SOURCES} ${COPY_FRAMEWORKS} libfmodex.dylib  /opt/local/lib/libpng16.16.dylib)
   #add_executable(editor MACOSX_BUNDLE OSX/SDLmain.m ${EDITOR_SOURCES} libfmodex.dylib  /opt/local/lib/libpng16.16.dylib)
   if (FMOD_ENABLED)
-    add_executable(editor MACOSX_BUNDLE ${EDITOR_SOURCES} libfmodex.dylib  /opt/local/lib/libpng16.16.dylib)
+	  add_executable(${EDITOR_EXE_NAME} MACOSX_BUNDLE ${EDITOR_SOURCES} libfmodex.dylib  /opt/local/lib/libpng16.16.dylib)
```
lheckemann: The indentation should be all spaces or all tabs, let's not mix them. My preference would be 2 spaces as this seems to be the most common standard for cmake, but I don't care that much as long as it's consistent. @addictgamer ?

CMakeLists.txt
```diff
@@ -353,91 +356,112 @@ if (OPENAL)
...
-  target_link_libraries(editor -lSDL2)
-  target_link_libraries(editor -lSDL2_net)
-  target_link_libraries(editor -lSDL2_image)
-  target_link_libraries(editor -lSDL2_ttf)
+	target_link_libraries(${EDITOR_EXE_NAME} -lmingw32)
+	target_link_libraries(${EDITOR_EXE_NAME} -lSDL2main)
+	target_link_libraries(${EDITOR_EXE_NAME} -lSDL2)
+	target_link_libraries(${EDITOR_EXE_NAME} -lSDL2_net)
+	target_link_libraries(${EDITOR_EXE_NAME} -lSDL2_image)
+	target_link_libraries(${EDITOR_EXE_NAME} -lSDL2_ttf)
```
lheckemann: This could all be one `target_link_libraries` call.

CMakeLists.txt
```diff
@@ -353,91 +356,112 @@ if (OPENAL)
...
   else()
-    target_link_libraries(editor ${VORBISFILE_LIBRARY})
+	  target_link_libraries(${EDITOR_EXE_NAME} ${VORBISFILE_LIBRARY})
   endif()
 endif()
 
-#install(TARGETS editor
-#  RUNTIME DESTINATION bin COMPONENT Runtime
-#)
+if (NOT APPLE AND UNIX)
```
lheckemann: Any particular reason to exclude apple here? This could be useful for package managers, and the existence of an install target doesn't mean a developer has to use it if they'd rather not.

CMakeLists.txt
```diff
@@ -353,91 +356,112 @@ if (OPENAL)
...
+	install(TARGETS ${EDITOR_EXE_NAME}
+		RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR} COMPONENT Runtime
+	)
 
-#install(DIRECTORY data DESTINATION share/barony )
+install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/lang DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/barony)
```
lheckemann: indentation's missing here

CMakeLists.txt
```diff
@@ -353,91 +356,112 @@ if (OPENAL)
...
+
+	if (EXISTS books)
+		install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/books DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/barony)
+		install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/images DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/barony)
+		install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/items DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/barony)
+		install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/maps DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/barony)
+		install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/models DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/barony)
+		install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/music DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/barony)
+		install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/sound DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/barony)
```
lheckemann: this can all be one call to `install(DIRECTORY` — https://cmake.org/cmake/help/v3.3/command/install.html#installing-directories

CMakeLists.txt
devnexen: true I just lazily kept that way ... will correct all in a jiffy :-)

CMakeLists.txt
devnexen: This is the best I found so far not ideal maybe but that avoids using programmatically API not everywhere available to get the assets.

addictgamer:
Looks good, everything builds and works fine on my Linux and Mac machines.

Only thing I can think of is perhaps PATH_MAX should be defined in Config.hpp.in instead of main.hpp?

addictgamer:
> The indentation should be all spaces or all tabs, let's not mix them. My preference would be 2 spaces as this seems to be the most common standard for cmake, but I don't care that much as long as it's consistent. @addictgamer ?

I don't particularly care either, as long as it's consistent (oops!).
Seeing as 2 spaces does seem to be common practice, let's roll with that in the future @lheckemann

PR #275, opened 2017-12-19
Add -Wformat-security

lheckemann:
It's a sensible flag to add (my linux distro even adds it to builds by default) and isn't too hard to do.


devnexen:
While at default C/C++ flags, do you think flags like -fstack-protector* would be useful ? For the non pandora build at least (no idea what is pandora though :-))

lheckemann:
The [pandora](https://pyra-handheld.com/boards/pages/pandora/) is a handheld game console.

Yes, I think they'd make sense but only for debug builds (we don't want memory bugs that would otherwise not affect the game to crash it while end users are playing).

lheckemann:
Update: I retract what I previously wrote! Since Barony has networking support and as such untrusted input the stack protector **definitely** should be added! https://stackoverflow.com/a/4280418 explains decently.

addictgamer:
Compilation fails on a Mac:

```
/Users/ciprian/programming/turning_wheel/barony_opensource/src/main.hpp:26:10: fatal error: 
      'unordered_map' file not found
#include <unordered_map>
```

devnexen:
Odd it should work at worse you would get a warning if you tried to compile with "old" compiler. Tried locally a very basic hello world, compiled with both base clang and g++

addictgamer:
I don't think this flag in particular is the problem. Might have something to do with shuffling around the CMake build flags:
https://stackoverflow.com/a/40309252

I checked, and it looks like the Apple section did have `-stdlib=libc++` listed in it, so I'm not sure what the issue is.

lheckemann:
It looks like `-std=c++11` went missing in the process though!

addictgamer:
Seems so! Adding it back in didn't fix the issue though.

addictgamer:
Ah! I added ```set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -std=c++11 -stdlib=libc++")``` and now it works.

lheckemann:
It should go in `CMAKE_CXX_FLAGS`, not `CMAKE_CXX_FLAGS_DEBUG`, since it's a setting that applies regardless of the build type.

addictgamer:
> since it's a setting that applies regardless of the build type.

Hmm. Seems like it isn't. Adding it to CMAKE_CXX_FLAGS didn't fix the issue, but duplicating both of the flags into CMAKE_CXX_FLAGS_DEBUG did.

Perhaps my changes were getting overwritten further down in the file? I didn't check.

lheckemann:
OK so apparently we should *actually* be using [add_compile_options](https://cmake.org/cmake/help/latest/command/add_compile_options.html)… CMake is a mess

PR #283, opened 2017-12-24
Fix Issue #219 - Poison Kills Buddha

crkellen:
This is a fix for #219.
The issue is there are no checks for if Buddha mode is enabled when dealing damage from Poison Effect.
Additionally, there was a misspelling in the lang file, which was fixed.

### DIFF
--- a/lang/en.txt
+++ b/lang/en.txt
@@ -1037,7 +1037,7 @@ You should call a plumber!#
 637 Your %s flickers in the darkness.#
 638 Your %s burns out.#
 639 Your amulet glows a faint green.#
-640 The poison insantly subsides!#
+640 The poison instantly subsides!#
 641 The poison seems to have subsided.#
 642 You lose some blood.#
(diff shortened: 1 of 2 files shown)


addictgamer:
It doesn't seem I'm taking poison damage at all, buddha mode or otherwise?

crkellen:
I'm almost 100% certain I double checked and tested to be sure this time. Maybe I messed up and forgot something. I can fix it first thing in the morning.
null

crkellen:
I do not have those issues at all. I took damage from poison while in Buddha mode with health above 1, with Buddha mode with health at 1, without Buddha mode with health above 1, and without Buddha mode with health at 1.

crkellen:
I considered that as well. There was a reason I did not do it that way, and I believe it was because of visual reasons. I can't check right now, at work.
null

lheckemann:
@addictgamer does the fix work right in multiplayer?

PR #284, opened 2017-12-24
Fix Issue #220 - Hungry Buddha Dies

crkellen:
This is a fix for #220.
The issue is that there are no checks for if Buddha mode is enabled when processing damage from HUNGER effect.


crkellen:
Rebased to barony-next in #284.

crkellen:
Accidentally closed the wrong PR.

PR #286, opened 2017-12-25
Fix Issue #73 - Shopkeeper Hostility does not consistently trigger

crkellen:
This is a fix for #73.
The issue is that the code that triggers Shopkeeper's hostility towards Players is only performed in the "Charge" state. If a Player attacks a stationary Shopkeeper from up close, the Shopkeeper will move towards the Player, but will not trigger any aggression. However, if a Player is slightly further away, the Shopkeeper will move towards the Player, allowing the code to trigger, setting Player's as `swornenemies[SHOPKEEPER][HUMAN]`.

This is fixed by extracting the code out of the state machine. This allows for hostility to be triggered during any state. This does not fix the exact same issue for Humans (potential Followers and the like), which will be addressed in a separate PR.

*NOTE:* This will make Shopkeepers hostile against every single Human, not just Players. Players are not currently separated from Humans.


addictgamer:
The shopkeeper started attacking me after I opened and closed a trade screen with him.

crkellen:
Does the Shopkeeper set `monsterTarget` to the Player? Because that's how hostility is currently detected. It's been a major headache trying to track down all of the locations where `monsterTarget` is actually set, but the current implementation only detects Player hostility during the "Charge" state. My fix extracts it out of the State Machine, but doesn't change the setting of `monsterTarget`.

crkellen:
The solution to this issue requires a separation of `monsterTarget` into something more like `monsterTarget` and `monsterFocus` where the new `monsterFocus` is the Entity that the monster is currently focused on, rather than aggressively targeting.

If this separation can be made, I believe both this and #230 can be solved. As it currently stands, there are just too many cases for this to be elegantly solved. I could do it, but it would be an absolute mess, as the case where one Player is speaking to a Shopkeeper and another one hits the Shopkeeper is too difficult to easily cover.

Your thoughts on this?

PR #301, opened 2017-12-30
Creature list optimisation

addictgamer:
Refactored all the code to, instead of looping the entire entity list, loop only over creatures where applicable. This most noticeably affects monster AI (which calls it the most often, and the most times a second, and which has the most callers at any given time), but also things such as spells, arrows, traps, etc.

Unfortunately, while this will benefit singleplayer and servers, this probably won't do much for clients. Most of the client's equivalent time is spent in the netcode searching through the entire entity list to match entity UIDs it receives. That's a separate issue, oh boy.


addictgamer:
* Also added a couple of new commands.

PR #305, opened 2017-12-31
Limb dead reckoning fix

### LINKED ISSUE #300: Dead reckoning inefficient for player and monster limbs
Right now, the code looks like this:

In game.cpp:
```
			// run entity actions
			for ( node = map.entities->first; node != nullptr; node = nextnode )

...


									// dead reckoning

...

										if ( entity->behavior == &actPlayer )
										{
											for ( node2 = map.entities->first; node2 != nullptr; node2 = node2->next )

...

										}
										if ( entity->behavior == &actMonster ) //Same thing if I am a monster.
										{

```

So it loops through the entire entity list (fine, nothing we can do about that).
Then every time it hits a player or a monster, it reloops through the entire entity list! Searching for its limbs.

It should instead grab the limbs directly from itself -- no searching!

This will improve multiplayer performance.

lheckemann:
Should fix #300

### DIFF
--- a/src/actplayer.cpp
+++ b/src/actplayer.cpp
@@ -273,6 +273,7 @@ void actPlayer(Entity* my)
 			node->element = entity;
 			node->deconstructor = &emptyDeconstructor;
 			node->size = sizeof(Entity*);
+			my->bodyparts.push_back(entity);
 
 			// magic hands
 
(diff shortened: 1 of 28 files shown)


addictgamer:
Looks good to me

Issue #312, opened 2018-01-05
Clear versioning information


lheckemann:
Currently, the steam release is pretty much a black box to most people (I don't know if you have some out-of-band way of tracking this?). In any case, it would be nice to have a couple of measures in place which would make it easier to tell:

 - Consistent use of tags in the repo — tag everything that gets released to steam
 - Include abbreviated commit hash into version string (I'll probably prepare a PR for this)

Additionally, it would be nice if there were a way as a user who's bought the game to get the game data for an arbitrary version. I don't know of a way to get the data for any version other than the current one.

addictgamer:
> Consistent use of tags in the repo — tag everything that gets released to steam

This is something we need to be better about.
Any help would be appreciated.

lheckemann:
This isn't really something the community can help with — you're the ones who bake and push the releases and actually know which commits they correspond to.

addictgamer:
That's correct. The main thing I had in mind is I've made branches for specific releases before, but I've forgotten to tag the last commit on that branch as the release.

addictgamer:
At any rate, the v2.0.6 and v2.0.7 releases should now be tagged. I think we're up-to-date, at least until we officially launch the Blessed Addtion.

lheckemann:
Yay!

It would be nice to have something like v3.0.0preXYZ tags for the beta versions pushed to steam as well.

addictgamer:
Ya, I don't even know where we stand with those ones. I believe the public-beta-next is representative of what's currently on the Steam version? Anything we may have had live before that though (if anything), it's too late now.

I'll go ahead and push a tag for that at least.

addictgamer:
Ok, everything should now be as up-to-date as we can make it right now.

Remind me if I forget to push tags for the beta and launch :)

lheckemann:
Thanks!

sebleblanc:
v3.3.3, v3.3.2 and v3.3.1 do not have tags in the repository.

v3.3.3: b70f35d9b8c89e43265ddd0fe528de6867b1eece
v3.3.2: 5cc7bfe53ba62ccfb0d0b5665546fd201ccfd235
v3.3.1: 2acc68c6fe953a46f5738d4b9279af3cbb2ca2ed
v3.3.0: 2acc68c6fe953a46f5738d4b9279af3cbb2ca2ed
v3.2.5: 53948cbf8de9ef259123e6ad89ea163f01a388e4
[...]

PR #313, opened 2018-01-05
Include short git hash in version string

lheckemann:
This can make tracing issues back to the code a lot easier, especially
with steam versions, for people who do not know what happens to be the
current version on Steam etc., see #312


addictgamer:
I'd rather the git hash and the version number be separate strings, then draw them together in the menus.

lheckemann:
You're right, this breaks scores handling as is. Will fix.

PR #318, opened 2018-01-10
getLinesFromFile -> getLinesFromDataFile

### LINKED ISSUE #307: Artifacts are too common
On a recent run, I obtained:

- 7× Dyrnwyn
- 3× Sharur
- 3× Gungnir
- 1× Khryselekatos

I think they should maybe be a *little* rarer.

EDIT: Also multiple artifact helms/breastplates/cloaks/etc

lheckemann:
This has several benefits, especially with regard to support for
alternative data directories:

 - Supports datadir without trailing slash for player name lists
 - Fixes #307
 - Reduces assumptions about datadir outside files.cpp

### DIFF
--- a/src/cppfuncs.hpp
+++ b/src/cppfuncs.hpp
@@ -27,25 +27,3 @@ T randomEntryFromVector(std::vector<T> vector)
 
 	return vector[rand() % vector.size()];
 }
-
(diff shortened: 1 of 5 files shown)


addictgamer:
Looks good.

PR #319, opened 2018-01-12
Editor improvements

lheckemann:
Sorted filenames (makes finding the map you're looking for so much easier!) and fix crashing when using a different datadir.


addictgamer:
Works in Linux and Mac.

PR #321, opened 2018-01-23
Updated/extended contributing and install information.

kornl:
For an even shorter summary then the short information below:
- Info about Discord channel for contributing
- Added links
- Info about libz used by libpng
- FMOD deprecated (and disabled if not found)
- Link to GL header files


addictgamer:
Looks good

PR #322, opened 2018-01-25
Misc balance and additions

WALLOFJUSTICE:


src/game.cpp
```diff
@@ -2270,11 +2270,20 @@ int main(int argc, char** argv)
...
-			                         "Barony has encountered a critical error and cannot start.\n\n"
-			                         "Please check the log.txt file in the game directory for additional info,\n"
-			                         "or contact us through our website at http://www.baronygame.com/ for support.",
-			                         screen);
+									"Barony has encountered a critical error and cannot start.\n\n"
+									"Please check the log.txt file in the game directory for additional info\n"
+									"and verify Steam is running. Alternatively, contact us through our website\n"
+									"at http://www.baronygame.com/ for support.",
+				screen);
+#elif
```
addictgamer: Should be `#else`?

WALLOFJUSTICE: Addressed in 51b5ccb

src/actgeneral.cpp
```diff
@@ -471,4 +471,8 @@ void actFloorDecoration(Entity* my)
 	{
 		return;
 	}
+	if ( !my->flags[PASSABLE] )
+	{
+		my->flags[PASSABLE] == true;
```
addictgamer: This code is incorrect

addictgamer: This may also be why pathfinding was breaking with grass

addictgamer:
Merged.

PR #327, opened 2018-03-10
Two fixes : $ORIGIN was considered as a variable and

devnexen:
was not used since it s a linker flag


addictgamer:
Hmm, not sure about this. There's probably some leftover cruft back from when I was figuring out Mac's "dynamic" library loading (hardcoded "install" paths -- bleh!). Unfortunately I don't have an easy way of testing Mac binaries because the only Mac I have access to is mine, and it's a dev machine (everything always works perfectly fine on a dev machine).

devnexen:
-Wl,-rpath is definitely a linker flag but it s not a big deal, we can leave things as it is, actually just it throws lot of warning as 
"...-Wl,-rpath=$RIGIN not used" (as you can see the O is missing) for each object compilation.

PR #331, opened 2018-03-15
Multiple save/score files, misc fixes

WALLOFJUSTICE:


src/sound_game.cpp
```diff
@@ -397,6 +399,11 @@ void handleLevelMusic()
 			activeminotaur = true;
 			break;
 		}
+		else if ( entity->sprite == 646 || entity->sprite == 650 )     // magister body
```
addictgamer: Could we transition this whole section to use map.creatures and check for races instead of sprites? That would be way more efficient.

addictgamer:
Regarding the first comment, I'm going to push my temporary fix. Follow up with whatever additional changes might be necessary.

PR #341, opened 2018-03-19
Sdl2 constant fixes and buffer overflow fix

devnexen:
1/ In the editor few messages were too long
2/ SDL2 now define HAVE_FMOD constant, renaming to fit with USE_TREMOR

### DIFF
--- a/src/Config.hpp.in
+++ b/src/Config.hpp.in
@@ -16,11 +16,11 @@
 #endif
 
 #if @FMOD@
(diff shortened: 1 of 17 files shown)


devnexen:
Just to give some clues, this is pretty recent in SDL2 as you can see here

http://lists.libsdl.org/pipermail/commits-libsdl.org/2017-November/002389.html

Just spotted when I updated to SDL 2.0.8

PR #344, opened 2018-03-23
Post-beta bugfix #3?

WALLOFJUSTICE:
+ Bunch of optimisations!


src/interface/identify_and_appraise.cpp
```diff
@@ -405,10 +405,43 @@ int getAppraisalTime(Item* item)
...
 	}
 	else
 	{
 		appraisal_time = (1000 * 60) / (stats[clientnum]->PROFICIENCIES[PRO_APPRAISAL] + 1);    // time in ticks until item is appraised+-
+		int playerCount = 0;
+		for ( int i = 0; i < MAXPLAYERS; ++i )
+		{
+			if ( !client_disconnected[i] )
+			{
+				++playerCount;
```
addictgamer: Might be useful to replace all these instances with a function?

PR #353, opened 2018-04-06
Mp port sanity

devnexen:
Since it s not an enterprise app, I was "nice" with the lower range :-)

### DIFF
--- a/src/game.hpp
+++ b/src/game.hpp
@@ -69,7 +69,7 @@ extern bool inrange[MAXPLAYERS];
 extern bool deleteallbuttons;
 extern Sint32 client_classes[MAXPLAYERS];
 extern Uint32 client_keepalive[MAXPLAYERS];
-extern Uint32 portnumber;
+extern Uint16 portnumber;
 extern list_t messages;
 extern list_t command_history;
 extern node_t* chosen_command;
@@ -191,7 +191,7 @@ void startMessages();
 void initClass(int player);
 
 extern char last_ip[64];
-extern char last_port[64];
+extern char last_port[6];
 
 //TODO: Maybe increase with level or something?
 //TODO: Pause health regen during combat?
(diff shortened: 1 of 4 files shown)


addictgamer:
@devnexen  Looks like there's an issue in consolecommand.cpp when reading in a port?

```
	else if (!strncmp(command_str, "/port ", 6))
	{
		if (command_str[6] != 0)
		{
			strcpy(last_port, command_str + 6);
			last_port[strlen(last_port) - 1] = 0;
		}
	}
```

Given the attached config file, this crashes on the current steam binary for version 3.1.4.
[default.cfg.txt](https://github.com/TurningWheel/Barony/files/1914333/default.cfg.txt)

If users delete their default.cfg, that resolves the problem (at least until they try to host/join a new directip game and a new port is saved).

devnexen:
I see ... maybe puttng back last_port to 64 length will solve this ?

addictgamer:
```
==23224==ERROR: AddressSanitizer: global-buffer-overflow on address 0x55821c28a906 at pc 0x7f0c2250ac91 bp 0x7ffc28ad4950 sp 0x7ffc28ad40f8
WRITE of size 7 at 0x55821c28a906 thread T0
    #0 0x7f0c2250ac90 in __interceptor_strcpy /build/gcc/src/gcc/libsanitizer/asan/asan_interceptors.cc:546
    #1 0x55821a5dfebb in consoleCommand(char*) /programming/turning_wheel/barony_opensource/src/interface/consolecommand.cpp:807
    #2 0x55821a5c7aa6 in loadConfig(char*) /programming/turning_wheel/barony_opensource/src/interface/interface.cpp:806
    #3 0x558219faf8ca in main /programming/turning_wheel/barony_opensource/src/game.cpp:2290
    #4 0x7f0c1f449f49 in __libc_start_main (/usr/lib/libc.so.6+0x20f49)
    #5 0x558219dffe69 in _start (/programming/turning_wheel/barony_content/barony+0x11d7e69)
```

Ya, I guess just increasing the size should do it. Looks like it's trying to write 7 into an array of size 6.

devnexen:
Sorry I did not spot as I use always new config, new savegame ... etc each time :-|

PR #359, opened 2018-04-15
use after free when end of party

devnexen:
not urgent to merge, showing up only using compiler sanitisers.


devnexen:
Any chance before the next release ? :-)

PR #370, opened 2018-05-04
Folder for datafiles

devnexen:


devnexen:
Unixes users reported that datafilee are created in the current folder which can be an issue most of the time. So here a proposal to create a sane place to store those.

devnexen:
Any chance before the next "Gog" release ? :-)

WALLOFJUSTICE:
I'll try sift this through soon, I've made some changes to save file names and such so will be a little manual work to check.

If you want to have a quick check for other areas that would require this fix for unix my development branch is here https://github.com/WALLOFJUSTICE/Barony/tree/dev-achievements

devnexen:
Seem your work is related to steam workshop in first glance.

devnexen:
I had a look and seemingly the only real conflict occurs in a pure steam workshop code part but I m not sure it s an issue in this case as the file saving issue happen for players with non packaged version of the game. There is a little need to adapt files.hpp

devnexen:
Unrelated note, maybe you can optimise a little this line for example
	https://github.com/WALLOFJUSTICE/Barony/blob/988d76b0825535421879c7f70e4f5d75a5484838/src/scores.cpp#L3344

devnexen:
I committed last changes, let me know if it works ok for you.

WALLOFJUSTICE:
Thanks this looks like it solves the rest of the issues, will test on Linux with Steam mods

PR #379, opened 2018-06-23
* Finish workshop support, QoL + Bugfixes

WALLOFJUSTICE:


src/files.cpp
```diff
@@ -1303,7 +1303,8 @@ void physfsReloadSounds(bool reloadAll)
 				{
 #ifdef USE_FMOD
 					FMOD_Sound_Release(sounds[c]);    //Free the sound in FMOD
-#elifdef USE_OPENAL
+#endif
+#ifdef USE_OPENAL
```
addictgamer: Possible issue here: both USE_FMOD and USE_OPENAL defined.

addictgamer: In hindsight, we should have been using #if & co instead of #ifdef & co.

PR #385, opened 2018-08-10
the array is one element short MAPFLAGS = MAP_FLAGS_DISABL…

devnexen:
…EOPENING


WALLOFJUSTICE:
Good spot, thanks! I'm going to leave the MAPFLAGS as is since I don't reference the map.flags[] array with the opening constant, only the char mapflagtext[MAPFLAGS][32] array for printing. Will make a MAPFLAGTEXT = 17 constant for this.

devnexen:
Sure. What works best for you :-)

PR #388, opened 2018-08-16
Fixing FPE when an item is not initialised.

devnexen:


devnexen:
Spotted first with 3.1.4 version but thought it was due to missing data but then occurred again (but randomly and rarely) with 3.2.0 as well, basically 216 items while possibly 218.

WALLOFJUSTICE:
To clarify, 3.2.0 increased the number of items by 1, as well as adding a new entry to items.txt. Likely crash is  on a line like (appearance % items[type].variations) where variations is uninitialised and could be 0. 

Solution can also be just to properly initialise some item properties to non-0

devnexen:
Yes exactly a fpe is division per zero.

PR #398, opened 2018-09-02
Use tile entity buckets, use high resolution frameRateLimit()

WALLOFJUSTICE:
+ Add option to disable steam multithreading
* Add tooltips on 'misc' settings tab options for clarity


addictgamer:
I get muuuuuch better performance now! :+1:

Issue #289, opened 2017-12-26
[Request] Add current Hunger status to the Character Sheet


crkellen:
This is a request from scayman223 on the Discord, written in my words.

There is a feature of unblessed mirrors that gives you a message hinting towards how hungry you currently are. By itself this is not an issue, but that is the only way of knowing your current hunger status other than mentally keeping track or being low enough that the icon appears.

There are cases when a Player may have spare food in their inventory, and would prefer to just maximize efficiency by eating it instead of dropping it. However, if you eat too much, you will become sick. This creates a need for there to be some way of knowing how full you are.

I propose that some sort of tracker be added to the TAB menu, perhaps on the Character Sheet, or somewhere else. Whether it is a bar, number, or even a short phrase or word, some sort of indicator is needed.

Armaly:
#419 

This is what i've done so far. Let me know if you would like anything changed.

PR #419, opened 2019-05-10
Fix Issue #289 - Added current hunger status to character sheet

Armaly:
-Modified updatecharactersheet.cpp to accomodate the menu's new stat.
-Increase lang array's size to accomodate the hunger stat in en.txt.

See #289

### DIFF
--- a/src/interface/updatecharactersheet.cpp
+++ b/src/interface/updatecharactersheet.cpp
@@ -54,7 +54,7 @@ void updateCharacterSheet()
 	//pos.w=222; pos.h=392-196;
 	//drawTooltip(&pos);
 	int statWindowY = 196;
-	int statWindowY2 = 404;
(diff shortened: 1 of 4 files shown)


src/interface/updatecharactersheet.cpp
```diff
@@ -1576,13 +1581,6 @@ void statsHoverText(Stat* tmpStat)
 							{
 								tmp = players[clientnum]->entity;
 								real_t regen = (static_cast<real_t>(tmp->getHealthRegenInterval(*tmpStat)) / TICKS_PER_SECOND);
-								if ( tmpStat->type == SKELETON )
```
WALLOFJUSTICE: Curious why this code got deleted? may need to pull this file again.. this is to show the regen values of a player skeleton.

PR #425, opened 2019-11-14
Replaced character classes if statements with json to make it more modular.

beef331:
Changed how character data is stored and loaded, to allow more moddabillity, and also to clean the code base up. Have not touched DLC characters.


WALLOFJUSTICE:
Thanks beef we'll be looking into pulling JSON integration shortly as we've just released the latest big update we've been working on.

PR #436, opened 2020-01-12
JSON file reader/writer

SheridanR:
Adds files to read/write json to/from classes and structs. Process is pretty simple, there's an example in json.hpp:

```
	class ExampleClass {
	private:
		Uint32 MyNumber;
		String MyString;
	public:
		// defines how to read/write the file to json:
		void serialize(FileInterface * file) {
			file->property("MyNumber", MyNumber);
			file->property("MyString", MyString);
		}
	};
```

Then when you want to read a file into an object:

```
// read an object from a json file:
ExampleClass exampleObject;
FileHelper::readObject("myJsonFile.json", exampleObject);
```

Or write a file from an object:

```
// write an object to a json file:
ExampleClass exampleObject;
FileHelper::writeObject("myJsonFile.json", EFileFormat::Json, exampleObject);
```

The benefits of this are huge. JSON is a standard and widely popular format. You can serialize either text-based JSON or binary json (text-based can be edited with a text editor, binary is significantly faster for serializing). The game will not crash when attempting to read missing properties. Versioning is trivial (simply add a "version" property in the serialize() function, as so):

```
		void serialize(FileInterface * file) {
			int version = 2;
			file->property("version", version);
			if (version < 2) {
				// loading an old version
			} else {
				// loading the latest version
			}
		}
```

This PR adds a new dependency to rapidjson and so modifies the LICENSE.txt to include a blurb about it.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -65,3 +65,20 @@ freely, subject to the following restrictions:
    misrepresented as being the original software.
 3. This notice may not be removed or altered from any source distribution.
 
+
+
+RapidJSON License:
+
+Tencent is pleased to support the open source community by making RapidJSON available.
+
+Copyright (C) 2015 THL A29 Limited, a Tencent company, and Milo Yip. All rights reserved.
+
+Licensed under the MIT License (the "License"); you may not use this file except
+in compliance with the License. You may obtain a copy of the License at
+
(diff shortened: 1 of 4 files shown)


addictgamer:
Does this have everything required to build, such as the CMake module?

PR #440, opened 2020-01-24
Added Linux code for log rotation

addictgamer:
Please verify this still works in Windows.


addictgamer:
Verified it works on Mac.

PR #457, opened 2020-04-01
Remove GLUT dependency

glebm:
GLUT is deprecated and is no longer available on many platforms.


SheridanR (approved):
nice work! glut is a terrible API, I frankly forgot we still had that as a dependency.

addictgamer:
Ah, so those merge conflicts: I just noticed it looks like also we got rid of this in the Nintendo port (and the changes are available back in master due to the check-in): https://github.com/TurningWheel/Barony/commit/3d7f914203ab5054bb96b043e94df983821e3f02#diff-70c7e5bff06780c55ac95be38b188c06a1120ccb293d0bf2d6145281917dca35R29

I guess all that remains is cleaning up the build files.

addictgamer:
Merged in cleanup of the remaining references (CMakeLists and a header include) in #562.

PR #471, opened 2020-06-13
Cleaning Code - items

DrWhoCares:
### Added:
- Added copy and move operations to SummonProperties to adhere to the Rule of 5.
---
### Modified:
- Replaced C-style casts with static_cast.
- Replaced NULL with nullptr.
- Joined declaration with assignment where possible.
- Modified shouldInvertEquipmentBeatitude() so that the parameter is const (as it should be).
- Replaced traditional for loops with ranged-based for loops.
- Made member functions const where possible.
- Fixed two instances of int being used instead of Uint32.
- Replaced while(1) with while(true).
- Made parameters const where possible.
- Made local variables const where possible.
- Replaced remaining instances of C-style casts with reinterpret_cast.
- Used auto instead of duplicating type in places that are static_casted.
- Made SummonProperties constructor and destructor default, and the destructor noexcept.
- Initialized 'itemId' to -1 in createCustomInventory() to prevent undefined behavior.
- Fixed 'itemToFind' parameter of itemNodeInInventory() to be an int32_t instead of ItemType, so -1 is a more defined option.
- Removed unreachable code and fixed indentation in items_editor itemSprite().
---
### Removed:
- Removed redundant else statements.
- Removed unreachable code.
- Removed redundant (harmless) control statements.
- Commented out completely unused and/or unimplemented code in items_editor.cpp newItem().


src/items.cpp
```diff
@@ -1273,93 +1259,92 @@ bool dropItem(Item* item, int player, bool notifyMessage)
 		}
 		return false;
 	}
-	else
```
WALLOFJUSTICE, line 1276: Personal taste - prefer the else {} as it indicates the below code is strictly for singleplayer/server

src/items.cpp
```diff
@@ -1205,12 +1191,12 @@ bool dropItem(Item* item, int player, bool notifyMessage)
 
 	if ( multiplayer == CLIENT )
 	{
-		strcpy((char*)net_packet->data, "DROP");
```
WALLOFJUSTICE, line 1208: I'd rather leave as c-style casts for all the network packet code due to quantity of changes will be a lot

src/items.cpp
```diff
@@ -233,23 +231,22 @@ Item* newItem(ItemType type, Status status, Sint16 beatitude, Sint16 count, Uint
 		{
 			if ( inventory == &stats[clientnum]->inventory )
 			{
-				int c;
```
WALLOFJUSTICE, line 236: Not a fan of refactoring hotbar[] to auto - we will be revisiting the hotbar code for splitscreen so I'd rather leave it all in the same intact way. I have manually changed it back after the charclass PR

src/items.cpp
```diff
@@ -105,12 +105,12 @@ Item* newItem(ItemType type, Status status, Sint16 beatitude, Sint16 count, Uint
 		}
 		else if ( multiplayer != CLIENT )
 		{
-			for ( int i = 0; i < MAXPLAYERS; ++i )
```
WALLOFJUSTICE, line 108: Would rather auto only be used for std containers such as vectors/lists instead of array[]

src/items.cpp
```diff
@@ -1058,9 +1047,9 @@ SDL_Surface* itemSprite(Item* item)
 			node_t* node = list_Node(&items[item->type].surfaces, spell->ID);
 			if ( !node )
 			{
-				return NULL;
+				return nullptr;
 			}
-			SDL_Surface** surface = (SDL_Surface**)node->element;
+			auto** surface = static_cast<SDL_Surface**>(node->element);
```
WALLOFJUSTICE, line 1052: unnecessary auto, recommend leave as SDL_Surface**

src/items.cpp
```diff
@@ -1069,12 +1058,12 @@ SDL_Surface* itemSprite(Item* item)
 		node_t* node = list_Node(&items[item->type].surfaces, item->appearance % items[item->type].variations);
 		if ( !node )
 		{
-			return NULL;
+			return nullptr;
 		}
-		SDL_Surface** surface = (SDL_Surface**)node->element;
+		auto** surface = static_cast<SDL_Surface**>(node->element);
```
WALLOFJUSTICE, line 1063: unnecessary auto, recommend leave as SDL_Surface**

src/items.cpp
```diff
@@ -1273,93 +1259,92 @@ bool dropItem(Item* item, int player, bool notifyMessage)
...
 			}
-		}
-		else
-		{
-			item->count = item->count - qtyToDrop;
+			item->count = oldcount - qtyToDrop;
 			if ( item->count <= 0 )
 			{
-				free(item);
+				list_RemoveNode(item->node);
```
WALLOFJUSTICE, line 1329: This line shows how difficult PRs are sometimes, the PR is correct but github shows this changing a line earlier in the code.. would have been a breaking change. This is fine

src/items.cpp
```diff
@@ -1606,13 +1591,12 @@ void consumeItem(Item*& item, int player)
 	{
 		if ( item->node != nullptr )
 		{
-			int i;
-			for ( i = 0; i < MAXPLAYERS; i++ )
+			for ( auto* const stat : stats )
```
WALLOFJUSTICE, line 1594: again, should be left as standard for loop instead of auto

src/items.cpp
```diff
@@ -1753,89 +1737,86 @@ EquipItemResult equipItem(Item* item, Item** slot, int player)
 		}
 		return EQUIP_ITEM_SUCCESS_NEWITEM;
 	}
-	else
```
WALLOFJUSTICE, line 1756: again the else {} separates scope from the above as preference

src/items.cpp
```diff
@@ -4707,57 +4665,67 @@ bool itemSpriteIsBreastpiece(int sprite)
 	return false;
 }
 
-real_t getArtifactWeaponEffectChance(ItemType type, Stat& wielder, real_t* effectAmount)
+real_t getArtifactWeaponEffectChance(const ItemType type, Stat& wielder, real_t* const effectAmount)
 {
 	if ( type == ARTIFACT_AXE )
 	{
-		real_t percent = 25 * (wielder.PROFICIENCIES[PRO_AXE]) / 100.f; //0-25%
+		const real_t percent = 25 * (wielder.PROFICIENCIES[PRO_AXE]) / 100.f; //0-25%
```
WALLOFJUSTICE, line 4672: adding const to variables that aren't function parameters isn't necessary

src/items.cpp
DrWhoCares: Isn't it the double explicit typing that's unnecessary though?
For example, if you needed to change this from being `SDL_Surface**` to some other type, you now need to change both instances.
I'm highly for explicit typing in almost every single case, but I do find that this case using auto is better, as it reduces overhead costs in development type if this ever needs to change.

src/items.cpp
DrWhoCares: Fixed in f76c8571

src/items.cpp
DrWhoCares: Fixed in f76c8571

src/items.cpp
DrWhoCares: Fixed in 7b658c6c

src/items.cpp
DrWhoCares: Fixed in ba4ea3fb

src/items.cpp
DrWhoCares: Very odd why it says something changed there.

src/items.cpp
DrWhoCares: Fixed in f76c857

src/items.cpp
DrWhoCares: This will be kept in mind for future commits, except in cases where it is important (which I don't foresee coming up often). If you prefer I undo the changes in this commit, let me know, otherwise, going forward.

src/items.cpp
DrWhoCares: Fixed in cd73a285

DrWhoCares:
I somehow missed that there was a comment at the beginning from you.
`Would like to see less or 0 rearranging of if-statement nesting`
So I noticed this after.
Do you want me to revert the commit which removed redundant else statements entirely?
I manually reverted the two instances you pointed out, And going forward, I'll skip that step unless important (such as hiding a bug). Let me know if you want me to revert it.

Lastly just need a response on my comment about the "using auto when explicitly casting" and then this should be set.

WALLOFJUSTICE:
Yes, please revert the else statements. As for the auto, change back to SDL_Surface** just for consistency in the codebase elsewhere.

DrWhoCares:
Reverted the rest of the removed redundant else statements in c3450de8

src/items.cpp
DrWhoCares: Fixed in 3010621c

src/items.cpp
DrWhoCares: Fixed in 3010621c

PR #520, opened 2020-08-20
GCC warning cleanup and some fixes

PegasusEpsilon:
Enjoy.


src/list.cpp
```diff
@@ -113,7 +113,7 @@ void list_RemoveNode(node_t* node)
 	}
 	else
 	{
-		delete (node->element);
+		free(node->element);
```
addictgamer, line 116: Good catch. Delete also shouldn't be used against malloc.

src/json.cpp
```diff
@@ -6,7 +6,7 @@
 #include "rapidjson/prettywriter.h"
 #include "rapidjson/error/en.h"
 
-const Uint32 BinaryFormatTag = 'spff';
+const Uint32 BinaryFormatTag = *"spff";
```
addictgamer, line 9: This was actually technically legal (although StackOverflow warns the reader about portability)

PegasusEpsilon: Legal or not, my objective was to get GCC to shut up when compiling a release build. Aside from the explosion from the sound engine not being done getting replaced (comment lines 2828 to 2851 in interface/consolecommand.cpp) it now compiles completely silently.

Debug builds are a different problem. I may or may not get around to cleaning that up, but there's a lot of work to do, there.

src/main.cpp
```diff
@@ -194,7 +194,7 @@ TCPsocket* net_tcpclients = nullptr;
 SDLNet_SocketSet tcpset = nullptr;
 list_t safePacketsSent, safePacketsReceived[MAXPLAYERS];
 bool receivedclientnum = false;
-char* window_title = nullptr;
+char const * window_title = nullptr;
```
addictgamer, line 197: Out of curiosity, any particular reason why this is being changed to `char const*`? I don't see what's wrong with letting the window_title string be modifiable (although I notice we don't ever directly modify it).

PegasusEpsilon: The C standard (I can't speak to C++) specifies that a string initialized that way can be stored in the code segment. GCC (I don't know about other compilers, but they probably do similar) is smart enough to move the modifiable string out of the code page.

The reason having a modifiable string in the code segment is bad is, if the OS and hardware support the NX bit, modifying it will cause a segmentation violation (SIGSEGV) and crash the program.

Intending to modify it or not, making it a const is The Right Thing.

There may be other instances where this error occurs, but this one gave me a warning, so I fixed it.

src/interface/interface.cpp
```diff
@@ -1359,10 +1356,7 @@ int saveConfig(char* filename)
 	}
 
 	fclose(fp);
-	if ( mallocd )
-	{
-		free(filename);
-	}
+	free(filename);
```
addictgamer, line 1359: Missing a couple other frees, such as on line 1098.

PegasusEpsilon: To be fair, that memory leak was there before me. I didn't do a full audit, the code has more issues than I really want to dive into today.

I did fix the const constraints, though, so that's progress.

addictgamer: Fair enough, that's true; can we touch up that lose end while we're here? ;)

PegasusEpsilon: Already done. 👉😉👉

addictgamer:
Looking at this PR makes me wish c-strings would leave our codebase and smart pointers would come in.

That's another story for another time, though. Thanks for the cleanup work!

PegasusEpsilon:
My thought while working on this is that I now understand why languages think all strings being immutable makes sense. It still doesn't, but at least I get it now.

The right fix to the problem, I think, is making char * = const char *const by default, and introducing a replacement keyword varying, so you have to specify VARYING char *VARYING instead.

But whatever, ISO never accepts my suggestions.

src/json.cpp
addictgamer: Understandable. I'm sad to admit the entire codebase is a royal mess, I'd rather not even think about debug builds right now.

src/main.cpp
addictgamer: So just for my own learning, is the issue coming from this pointer, or more from that it could potentially point to something like a string literal (which I believe actually may be the case)?

WALLOFJUSTICE:
Thanks for the PR!

DrWhoCares:
Personally, I think the C style variable definitions should've been changed to match the rest of the code base (which follows C++):
`char const * window_title` -> `const char* window_title`

However, this has already been merged, so it's more of an afterthought than anything.

PegasusEpsilon:
@DrWhoCares I thought about that halfway through editing the code, but it works fine either way, and automated tools can make that change (s/const char/char const/) if you really want it.

I personally object to a lot of formatting quirks in the code, but it's not my codebase.

DrWhoCares:
I don't disagree, but that's why I'd prefer to just follow the standards they've been attempting to maintain. To avoid adding more to the list of things to standardize later on.
Either way, I simply thought I'd voice my opinion on it. There's already plenty of outdated, and inconsistent code that this isn't a huge deal, especially when compared to the fixes the PR adds.

PegasusEpsilon:
No, I get it. It wasn't my intention to change the code style, I just didn't bother fixing the changes I'd already made when I noticed (halfway through) that you were using the other convention.

As for the fixes, there are a truckload of similar fixes left to be done. The Release build compiles silently, but the Debug build throws warnings all over the place. If I get super bored I may poke them, but definitely don't count on another episode of me tearing through the code, especially if it stops crashing on me after this PR.

If anyone else agrees (and decides they're likely to accept it) I'll throw a single "char const -> const char" patch for everything touched in this PR.

src/main.cpp
PegasusEpsilon: Yup. Points to a string literal, which may live in read-only memory, which when modified, would cause a segfault. Therefore we should tag it const.

PR #521, opened 2020-08-21
Relicensing Barony from GPL to BSD 2-Clause License

addictgamer:
Brief summary:
After some discussions within the team, we have decided to relicense to a more permissive license going forward to better suit our needs for Barony.
The PC version of Barony will remain open source; the only purpose of the more permissive license is to make it easier for us to port the game to proprietary platforms.

Our ask to you, our contributors:
 * Please sign off on this license change.
 * Should you object to this license change, we will be forced to revert your contributions.
 * If you do not respond by September 20, 2020, we will revert your commits.

Some additional background:
OpenAL has not been sufficient for our Nintendo Switch port, so we have had to go back to FMOD, which is incompatible with the GPL.
Additionally, due to the nature of our NDA with Nintendo, we are unable to publicly release some of the code for the Switch port. After internal discussions within the team, we've decided that switching to a more permissive license (in this case, the 2-Clause BSD license) is the best path for us going forward.

Contributors:
@DrWhoCares 
@crkellen 
@devnexen 
@lheckemann 
@ptitSeb 
@lakp 
@Gilgatex 
@pwhelan 
@kornl 
@Mas0s 
@Journeyman-dev 
@PegasusEpsilon


PegasusEpsilon:
Could you not make audio engine choice a runtime thing? I have no idea what FMOD looks like inside, or what other (non-linux) platforms look like for library loading, but...

Not that I object to a license change, but having more options makes me happy.

Like I'd be giddy if barony got a vulkan renderer. Ah, pipe dreams.

addictgamer:
> Like I'd be giddy if Barony got a vulkan renderer. Ah, pipe dreams.

I think a refactor to make Barony more modular would be a welcome change :+1: 
_(While we're on the subject, I think the sound engine is a great example of where we're lacking right now...you can't just swap OpenAL out for FMOD or vice versa, you have to make changes all over the codebase)_

> Could you not make audio engine choice a runtime thing? I have no idea what FMOD looks like inside, or what other (non-linux) platforms look like for library loading, but...

I'm sorry, not sure I follow?

DrWhoCares:
I have no issues with the update. I'm all for more permissive.

lakp:
Sure

PegasusEpsilon:
@addictgamer Like an in-game settings option, to use FMOD or OpenAL (or anything else, really) as the audio back-end.

In linux, you'd load your choice of shared object through dlopen() and then haul in the required function calls with dlsym(). You could even pass the work off to an external library that implements a "Barony audio API" wrapper for you, so your main code never has to worry about anything but what Barony actually needs, and how Barony actually needs it, and people can write audio layer wrappers around whatever driver they can make work with Barony. OSS, ALSA, OpenAL, FMOD, Pulse, Jack, anything.

Would also help with porting to switch, as you just write an audio wrapper for whatever the switch has, and never have to worry about closing part of the codebase, or changing the license. At least not for audio's sake.

addictgamer:
@PegasusEpsilon I see. Perhaps this discussion can be carried forward in a new issue? At any rate, I would not be opposed to such a change.

PegasusEpsilon:
@addictgamer I'd have to spend a week full-time going through the source to sufficiently understand all the touches that need to be made to disentangle the audio engine from the game proper, then design and implement a replacement that works with wrapper shims. And then there's the work to make it all portable. So yeah...I'm not about to build it myself, but it's an idea, at least.

devnexen:
+1 for me

ptitSeb:
I'm fine with the relicensing, but I wonder what is "not suffisant" with the OpenAL version..

brynet:
@addictgamer You mention that the switch to FMOD will allow ports to more proprietary platforms, but that happens at the expense of open platforms. As an example, FMOD is proprietary and unavailable for open source operating systems like *BSD. Switching to FMOD from OpenAL will negatively impact platforms that have already added Barony to their package repositories.

Speaking also as a non-contributor, but OpenBSD developer, I will add that while switching to a more permissive licence for Barony should have been a welcome direction appears to be not a positive one in terms of impact on the open source community. I also echo @lheckemann in saying that saying "if you do not respond by September 20, 2020, we will assume no objection with this license change." is legally dubious, if not totally illegal. This was similarly attempted by the OpenSSL project (not affiliated with OpenBSD) and was criticized. You must make an effort to contact those contributors, and if they do not reply, revert or rewrite their changes.

addictgamer:
To the best of our understanding, we need to move away from a copyleft license.
Barony is not architected in such a way that we can easily separate the free and non-free affected portions into separate components. After talking with the team, we consider the engineering effort that would be required is more than we're willing to commit to, so we are going forward with relicensing our codebase to the BSD license.

@lheckemann as for your contributions, it is up to you whether you accept this license change or, if you don't, we will comply with your decision and revert all of your commits.

Regarding the idea that someone might "take our code and run," I think it's worth mentioning that this would be against the hurdles of bad code written for a niche game using non-free assets. We don't believe there's substantial risk there, and as for us, we will continue to make all of our updates to the PC version available in this repository. I'd also like to make a note here: @sheridanr initially released Barony's code under the GPL for a reason -- we love copyleft and wish we could keep the code free in this manner, but given these practical problems it just is not workable.


@brynet to clarify, we are not removing OpenAL from the codebase. Barony will support both FMOD and OpenAL for its audio backend.
As for your second point, I have amended the description to explicitly state we will revert commits from any contributor which does not accept this license change. Thank you for your feedback.

pwhelan:
I give permission to relicense my contributions.

lheckemann:
Alright then, you have my permission.

Gilgatex:
I give permission to re-license

crkellen:
Yeah, I'm fine with the re-license.

kornl:
Yes, I'm also fine with the re-license and give permission for my neglectable work. ;) Thanks for everything!

Journeyman-dev:
Yes I give permission to re-license,

addictgamer:
It looks like we did not receive a response from @Mas0s.
I see one PR (https://github.com/TurningWheel/Barony/pulls?q=is%3Apr+author%3AMas0s), commit 507eefb7dcf5e3aded11d66b54405f177f206149 which I've reverted via d4a49c01a1f4ee6a9bd09d601b89d98fac0f8a40

addictgamer:
It looks like every other contributor has already responded in this PR, I'm going to go ahead and merge this now.
Thank you all.

PR #524, opened 2020-08-22
Haiku platform support.

devnexen:
Fixes in same time openal build.


CMakeLists.txt
```diff
@@ -403,7 +406,12 @@ if (GAME_ENABLED)
   endif()
   target_link_libraries(barony ${OPENGL_LIBRARIES})
   target_link_libraries(barony ${THREADS_LIBRARIES})
-  target_link_libraries(barony -lm -lc)
+  target_link_libraries(barony -lm)
+  if (NOT CMAKE_SYSTEM_NAME MATCHES "Haiku")
+	  target_link_libraries(barony -lc)
+  else()
+	  target_link_libraries(barony -lroot)
```
extrowerk, line 413: Unneeded, libroot will be unconditionally linked in.

CMakeLists.txt
```diff
@@ -502,7 +510,12 @@ if (EDITOR_ENABLED)
   endif()
   target_link_libraries(${EDITOR_EXE_NAME} ${OPENGL_LIBRARIES})
   target_link_libraries(${EDITOR_EXE_NAME} ${THREADS_LIBRARIES})
-  target_link_libraries(${EDITOR_EXE_NAME} -lm -lc)
+  target_link_libraries(${EDITOR_EXE_NAME} -lm)
+  if (NOT CMAKE_SYSTEM_NAME MATCHES "Haiku")
+	  target_link_libraries(${EDITOR_EXE_NAME} -lc)
+  else()
+	  target_link_libraries(${EDITOR_EXE_NAME} -lroot)
```
extrowerk, line 517: Also here.

cmake/Modules/FindSDL2_ttf.cmake
```diff
@@ -33,7 +33,7 @@ find_path(SDL2_TTF_INCLUDE_DIR SDL_ttf.h
   HINTS
     ENV SDLTTFDIR
     ENV SDLDIR
-  PATH_SUFFIXES include/SDL2 include
+    PATH_SUFFIXES include/SDL2 include headers/SDL2 headers
```
extrowerk, line 36: SDL2 should be enough

devnexen: You re surely right for the two points above but this particular point it did not work out of the box for me, is to avoid passing flags/env vars as much as possible.

extrowerk: headers folder always searched by cmake, and it will use the correct arch settings, if you run setarch. Also in haikuporter with correct prefix (_x86), so SDL header files can be found with only the SDL2 hint added here. See other ports for reference:
https://github.com/haikuports/haikuports/search?q=PATH_SUFFIXES+SDL&unscoped_q=PATH_SUFFIXES+SDL

devnexen: true you are correct I ve tried locally seems working finally.

PR #525, opened 2020-08-24
(Partial) Steam build vs GCC cleanup

PegasusEpsilon:
Fixed two potential segmentation violations, and one conversion of a pointer type to a non-pointer type.


src/menu.cpp
```diff
@@ -16182,11 +16182,15 @@ void buttonGamemodsStartUploadItem(button_t* my)
 	{
 		if ( gamemods_window == 5 )
 		{
-			g_SteamWorkshop->SubmitItemUpdate("Item updated.");
+			char *tmp = strdup("Item updated.");
+			g_SteamWorkshop->SubmitItemUpdate(tmp);
```
addictgamer, line 16186: I believe this API can be changed to `const char*` instead, that should resolve the issue right?

PegasusEpsilon: The API can be changed, yes, and I would also prefer that, but it currently makes no such promise. The bug would have to be reported upstream.

PegasusEpsilon: You are correct. I did this patch in 2 hours before a doctor's appointment right after I got up, and assumed SubmitItemUpdate was Steam API itself, rather than your wrapper. Proper patch pushed now.

PR #530, opened 2020-09-01
Removes unnecessary linkage for Haiku, does not have same curse as ot…

devnexen:
…her unixes

 in this context.
Pointed by extrowerk.


devnexen:
Not a big deal if not merged fir this release as it works as is already.

PR #539, opened 2020-09-29
Fix barony binary loading content.

devnexen:
Navigating through sub directories is related to Mac applications
 structure only.


addictgamer:
Is a similar change needed for the editor? If I recall correctly, I think we were doing similar chroot from editor.app into game.app...

PR #543, opened 2020-11-07
DragonFlyBSD support. It is just the same FreeBSD's code path.

devnexen:


CMakeLists.txt
```diff
@@ -398,7 +398,7 @@ if (GAME_ENABLED)
     if (APPLE)
       target_link_libraries(barony ${GLUT_LIBRARIES} ${PROJECT_SOURCE_DIR}/libpng16.16.dylib) #Wait...what? if(APPLE) in if(WIN32)? What was I thinking back then, haha.
     endif()
-    if (${CMAKE_SYSTEM_NAME} MATCHES "BSD" OR ${CMAKE_SYSTEM_NAME} STREQUAL "Haiku")
+    if (${CMAKE_SYSTEM_NAME} MATCHES "BSD|DragonFly" OR ${CMAKE_SYSTEM_NAME} STREQUAL "Haiku")
```
addictgamer, line 401: So, a question:
On line 145, I see we are checking if CMAKE_SYSTEM_NAME is FreeBSD, NetBSD, or DragonFly, but on this line we are checking if the same variable is BSD or DagonFly...shouldn't the check be the same on both lines, `${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD|NetBSD|DragonFly"`?

devnexen: Well the FreeBSD/NetBSD case is to discard OpenBSD which lacks support of code "self acknowlegment".

addictgamer: Oh, I see. So it is intentionally and specifically precluding OpenBSD...got it.
Since you're already there, could you also add a comment to that effect?
Thanks

addictgamer (approved):

PR #545, opened 2020-12-01
Fix a few straggling fopen() type code in mod_tools.cpp/mod_tools.hpp

SheridanR:

### DIFF
--- a/src/mod_tools.cpp
+++ b/src/mod_tools.cpp
@@ -174,15 +174,17 @@ void GameModeManager_t::Tutorial_t::readFromFile()
 		std::string inputPath = PHYSFS_getRealDir("/data/tutorial_strings.json");
 		inputPath.append("/data/tutorial_strings.json");
 
-		FILE* fp = fopen(inputPath.c_str(), "rb");
+		File* fp = FileIO::open(inputPath.c_str(), "rb");
 		if ( !fp )
 		{
 			printlog("[JSON]: Error: Could not locate json file %s", inputPath.c_str());
 			return;
 		}
 		char buf[65536];
-		rapidjson::FileReadStream is(fp, buf, sizeof(buf));
-		fclose(fp);
+		int count = fp->read(buf, sizeof(buf[0]), sizeof(buf));
+		buf[count] = '\0';
+		rapidjson::StringStream is(buf);
+		FileIO::close(fp);
 
 		rapidjson::Document d;
 		d.ParseStream(is);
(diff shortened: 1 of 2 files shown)


src/mod_tools.cpp
```diff
@@ -174,15 +174,17 @@ void GameModeManager_t::Tutorial_t::readFromFile()
...
 		if ( !fp )
 		{
 			printlog("[JSON]: Error: Could not locate json file %s", inputPath.c_str());
 			return;
 		}
 		char buf[65536];
-		rapidjson::FileReadStream is(fp, buf, sizeof(buf));
-		fclose(fp);
+		int count = fp->read(buf, sizeof(buf[0]), sizeof(buf));
+		buf[count] = '\0';
```
addictgamer, line 185: No check that count < buffer size?

src/mod_tools.cpp
```diff
@@ -215,15 +217,17 @@ void GameModeManager_t::Tutorial_t::readFromFile()
...
```
addictgamer, line 228: Likewise here...looks like there's a few other places like this.

src/mod_tools.hpp
```diff
@@ -1070,17 +1070,25 @@ class MonsterStatCustomManager
...
 		char buf[65536];
-		rapidjson::FileWriteStream os(fp, buf, sizeof(buf));
-		rapidjson::PrettyWriter<rapidjson::FileWriteStream> writer(os);
+		int count = fp->read(buf, sizeof(buf[0]), sizeof(buf));
+		buf[count] = '\0';
+		rapidjson::StringStream os(buf);
+		rapidjson::PrettyWriter<rapidjson::StringStream> writer(os);
 		d.Accept(writer);
+		for (int c = 0;;++c) {
+			if (buf[c] == '\0') {
```
addictgamer, line 1085: Wouldn't this check be equivalent if done in the for loop?
```for (int c = 0; buf[c] != '\0'; ++c)```

addictgamer (approved):
Looks good to me

src/mod_tools.cpp
SheridanR: Yeah I suppose that could be added, right now there's just an implicit expectation that the file will be <64kb. given they are just a few KBs right now, such an expectation seems reasonable...

addictgamer: Ya...one big thing I can think of is a malicious attack, but that isn't the biggest worry for this sort of program.

The other one is if we give jsons to modders to modify, we should have an explicit rejection of files they are making too big, rather than segfaulting (or worse: silently accepting the bad juju)

SheridanR: hold on, I went in to fix this and realized that fp->read() cannot overfill the buffer since it takes the size as the last parameter.

so the worst case scenario is that the json will be malformed, but that won't cause any serious error.

src/mod_tools.hpp
SheridanR: it would :) changed

src/mod_tools.cpp
addictgamer: Ah, that's good. A malformed JSON is much better than a hard crash :)

addictgamer: Ideally we'd still hit the user with a proper error so they know why their JSON is "malformed"...

PR #546, opened 2020-12-02
Misc fixes, splitscreen + nullptrs

WALLOFJUSTICE:


src/interface/updatechestinventory.cpp
```diff
@@ -78,9 +78,12 @@ int numItemsInChest()
 
 	int i = 0;
 
-	for (node = chestInventory->first; node != nullptr; node = node->next)
+	if ( chestInventory )
```
addictgamer, line 81: Aha, was that a chest crash we never chased down?

WALLOFJUSTICE: I never saw that one in action, but was definitely dangerous :D

src/magic/actmagic.cpp
```diff
@@ -3525,7 +3525,7 @@ void createParticleSap(Entity* parent)
 		}
 		entity->setUID(-3);
 
-		if ( sprite = 977 ) // boomerang
+		if ( sprite == 977 ) // boomerang
```
addictgamer, line 3528: Aha. I don't think we ever formally put this in our coding guidelines, but I think we've talked about putting rtypes on the left in equality checks...I don't like how they usually read, but hey, the compiler can catch typos like this for us.

PR #548, opened 2020-12-04
New UI engine

SheridanR:
A whole bunch of new classes from the Spacepunk engine that can be used to create more complex UIs, easier.

widgets:
Frame (window)
Field (text)
Button (checkbox, toggle, push)
Slider

Additionally, cached objects:
Text (rendered ttf strings)
Font (ttf)
Image


src/ui/Button.cpp
```diff
@@ -0,0 +1,204 @@
...
+	parent = &_parent;
+	_parent.getButtons().push_back(this);
+	_parent.adoptWidget(*this);
+}
+
+Button::~Button() {
+	if (callback) {
+		delete callback;
+		callback = nullptr;
+	}
```
addictgamer, line 28: How is cleanup initiated? No message needed to the parent; does the parent initiate the destruction of every button?

src/ui/Button.cpp
```diff
@@ -0,0 +1,204 @@
...
+		drawDepressed(_size.x, _size.y, _size.x + _size.w, _size.y + _size.h);
+	} else {
+		drawWindow(_size.x, _size.y, _size.x + _size.w, _size.y + _size.h);
+	}
+
+	if (!text.empty() && style != STYLE_CHECKBOX) {
+		Text* _text = Text::get(text.c_str(), font.c_str());
+		if (_text) {
+			SDL_Rect pos;
+			int textX = style == STYLE_DROPDOWN ? 5 + border : _size.w / 2 - _text->getWidth() / 2;
```
addictgamer, line 67: Minor suggestion: I think adding parentheses around the condition would help a bit with readability.

src/ui/Field.hpp
```diff
@@ -0,0 +1,98 @@
+//! @file Field.hpp
```
addictgamer, line 1: Maybe something cool we could do sometime is add decorators for these...actually, maybe all widgets could do with that.

src/ui/Slider.cpp
```diff
@@ -0,0 +1,189 @@
...
+		drawDepressed(_railSize.x, _railSize.y, _railSize.x + _railSize.w, _railSize.y + _railSize.h);
+	}
+	
+	// draw handle
+	_handleSize.x = _size.x + std::max(0, handleSize.x - _actualSize.x);
+	_handleSize.y = _size.y + std::max(0, handleSize.y - _actualSize.y);
+	_handleSize.w = std::min(handleSize.w, _size.w - handleSize.x + _actualSize.x) + std::min(0, handleSize.x - _actualSize.x);
+	_handleSize.h = std::min(handleSize.h, _size.h - handleSize.y + _actualSize.y) + std::min(0, handleSize.y - _actualSize.y);
+	if (_handleSize.w > 0 && _handleSize.h > 0) {
+		bool h = highlighted | selected;
```
addictgamer, line 35: What is h used for?

SheridanR: its a holdover from code i left out of spacepunk, it's just shorthand that would have been used in a branch to draw the element with more highlighted color but i've left out that bit for now.

src/ui/Frame.cpp
```diff
@@ -0,0 +1,1164 @@
...
+	Sint32 omousex = (omousex / (float)xres) * (float)Frame::virtualScreenX;
+	Sint32 omousey = (omousey / (float)yres) * (float)Frame::virtualScreenY;
+
+	if (selected) {
+		// unselect list
+		if (keystatus[SDL_SCANCODE_ESCAPE]) {
+			keystatus[SDL_SCANCODE_ESCAPE] = 0;
+			deselect();
+			if (!widgetBack.empty()) {
+				Frame* root = findSearchRoot(); assert(root);
```
addictgamer, line 409: Ah, are we building with asserts?

src/ui/Frame.cpp
```diff
@@ -0,0 +1,1164 @@
...
+			}
+		}
+	}
+
+	// process (widget) sliders
+	{
+		auto prev = sliders.rbegin();
+		for (auto it = sliders.rbegin(); it != sliders.rend(); it = prev) {
+			Slider* slider = *it;
+			prev = std::next(it);
```
addictgamer, line 668: So is this really a previous or is it more of a next?

src/ui/Frame.cpp
```diff
@@ -0,0 +1,1164 @@
...
+
+			++num;
+			prev = entry;
+		}
+	}
+
+	// process fields
+	{
+		auto prev = fields.rbegin();
+		for (auto it = fields.rbegin(); it != fields.rend(); ++it) {
```
addictgamer, line 750: Why do these 3 for loops look different?

	// process (widget) sliders
		for (auto it = sliders.rbegin(); it != sliders.rend(); **it = prev**) {

	// process the frame's list entries
		for (auto it = list.begin(); it != list.end(); **it = next**) {

	// process fields
		for (auto it = fields.rbegin(); it != fields.rend(); **++it**) {

src/ui/Frame.cpp
```diff
@@ -0,0 +1,1164 @@
...
+			delete slider;
+			sliders.erase(it);
+			return true;
+		}
+	}
+	return false;
+}
+
+bool Frame::removeEntry(const char* name, bool resizeFrame) {
+	entry_t* prevEntry = nullptr;
```
addictgamer, line 972: What is prevEntry used for?

addictgamer (approved):
CMake files also need to be updated.

Can't wait to see the new code in action :)

src/ui/Frame.cpp
SheridanR: std::next() for a reverse iterator is "previous" for a normal iterator :)

src/ui/Button.cpp
SheridanR: sure I can break it up a bit too and improve readability.

src/ui/Frame.cpp
addictgamer: Is this a reverse iterator? I see prev and iterator both start at rbegin, and prev seems to move on to the next item after it?

src/ui/Frame.cpp
SheridanR: I'm very fond of asserts as it explicitly draws the line between a working and broken state. there is no conceivable case in this code where you'd want root to be null. the assert implies that will never happens at runtime

src/ui/Frame.cpp
SheridanR: rbegin() returns a reverse iterator.

addictgamer: Also, I don't see `prev` being used anywhere. Why not do `++it` directly?

addictgamer: > rbegin() returns a reverse iterator.

Aha. I completely glossed over the r, oops.

src/ui/Frame.cpp
SheridanR: most ui elements are drawn one way and processed in reverse, so the last inserted element is most visible and also captures inputs first. this is not important for a frame list so it processes beginning to end.

src/ui/Frame.cpp
SheridanR: I don't embed it = std::next(it) in the for loop because the iterator could be invalid after the process runs (for example, a button that deletes itself). the next position is therefore saved so the increment is valid

SheridanR: actually i'm not sure how well tested that is since the next position could just as well be deleted, not to mention deletion happens in a later pass, so it might not be necessary

src/ui/Frame.cpp
addictgamer: Ya, I missed the rbegin/begin difference on my read-through, so I understand why the second looks different from the other two...

I guess my next question of why do 1 and 3 look different, would be answered by:
> I don't embed it = std::next(it) in the for loop because the iterator could be invalid after the process runs (for example, a button that deletes itself). the next position is therefore saved so the increment is valid

I'm assuming fields don't delete themselves (but sliders do), so we don't need to save the incrementor.

src/ui/Frame.cpp
SheridanR: Left that one in by mistake... I'll remove it

src/ui/Frame.cpp
addictgamer: Got it.

src/ui/Frame.cpp
addictgamer: Hmm, ok. I hope it never happens at runtime :)

addictgamer: (I usually just do checks for conditions over asserts, but even when I use asserts, I leave in some error handling code after the assert for the production builds...)

PR #550, opened 2020-12-06
splitscreen p1

WALLOFJUSTICE:


src/interface/interface.hpp
```diff
@@ -151,31 +149,31 @@ extern bool draw_cursor; //True if the gamepad's d-pad has been used to navigate
 void takeScreenshot();
 bool loadInterfaceResources();
 void freeInterfaceResources();
-void clickDescription(int player, Entity* entity);
+void clickDescription(const int player, Entity* entity);
```
addictgamer, line 152: Const for a non-pointer, non-reference type?

src/menu.cpp
```diff
@@ -10342,17 +10352,24 @@ void handleMainMenu(bool mode)
...
-			for ( c = 0; c < NUM_HOTBAR_ALTERNATES; ++c )
+
+			for ( int i = 0; i < MAXPLAYERS; ++i )
 			{
-				selected_spell_alternate[c] = NULL;
-				hotbarShapeshiftInit[c] = false;
+				deinitShapeshiftHotbar(i);
+				for ( c = 0; c < NUM_HOTBAR_ALTERNATES; ++c )
+				{
+					selected_spell_alternate[c] = NULL;
```
addictgamer, line 10361: We need to replace all of these `NULL`s with `nullptr`s at some point.

PR #552, opened 2020-12-15
UI elements scaled from virtual screen size to actual screen size

SheridanR:
Really simple PR that scales UI elements graphically from the virtual screen dimensions (1080p) to the real screen dimensions.

Check that I didn't mix any usage of xres/yres up! Some x + width / y + height properties are combined inline during scaling.

### DIFF
--- a/src/ui/Frame.cpp
+++ b/src/ui/Frame.cpp
@@ -123,7 +123,12 @@ void Frame::draw(SDL_Rect _size, SDL_Rect _actualSize) {
 
 	// draw frame background
 	if (!hollow) {
-		drawRect(&_size, color, (Uint8)(color>>mainsurface->format->Ashift));
+		SDL_Rect scaledSize;
+		scaledSize.x = _size.x * (float)xres / (float)Frame::virtualScreenX;
+		scaledSize.y = _size.y * (float)yres / (float)Frame::virtualScreenY;
+		scaledSize.w = _size.w * (float)xres / (float)Frame::virtualScreenX;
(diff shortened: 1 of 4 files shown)


src/ui/Frame.cpp
```diff
@@ -179,20 +189,26 @@ void Frame::draw(SDL_Rect _size, SDL_Rect _actualSize) {
...
 		handleRect.h = handleSize;
+
+		int x = (handleRect.x) * (float)xres / (float)Frame::virtualScreenX;
+		int y = (handleRect.y) * (float)yres / (float)Frame::virtualScreenY;
+		int w = (handleRect.x + handleRect.w) * (float)xres / (float)Frame::virtualScreenX;
+		int h = (handleRect.y + handleRect.h) * (float)yres / (float)Frame::virtualScreenY;
 		if (rectContainsPoint(barRect, omousex, omousey)) {
-			drawWindow(handleRect.x, handleRect.y, handleRect.x + handleRect.w, handleRect.y + handleRect.h);
+			// TODO highlight
+			drawWindow(x, y, w, h);
```
addictgamer, line 199: so w and h are really like x2 and y2, where x2 = x + w right?

addictgamer (approved):

addictgamer:
Looks to me like all of your xywh's are in order, I didn't spot anything off.

src/ui/Frame.cpp
SheridanR: yeah I haven't namely a variable x2 or y2 in so long I didn't think to use it!

PR #563, opened 2021-01-02
Fix OpenAL segfault in hamlet

Jamesernator:
This fixes a crash in hamlet when using a binary compiled for OpenAL that is caused by the `NUMMINESMUSIC` being different in OpenAL to FMOD ([the fmod constant is here](https://github.com/TurningWheel/Barony/blob/master/src/sound.hpp#L55)).

I also checked the other constants, and `NUMSWAMPMUSIC` is also wrong, but this doesn't lead to a segfault as the track just winds up unused instead. I've fixed this one as well so that the track can play when randomly chosen.

### DIFF
--- a/src/sound.hpp
+++ b/src/sound.hpp
@@ -142,9 +142,9 @@ struct FMOD_VECTOR {
 extern OPENAL_BUFFER** sounds;
 extern Uint32 numsounds;
 extern OPENAL_BUFFER** minesmusic;
-#define NUMMINESMUSIC 4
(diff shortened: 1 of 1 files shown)


addictgamer:
Sometimes I need brain bleach after reading old code.

Two questions for my past self:
1) Why is that copypasta? The backend doesn't matter for this.
2) Why isn't that dynamically figured out by just parsing the directory and allocating the proper array/list/whatever?

WALLOFJUSTICE:
Thanks for the fix!

PR #564, opened 2021-01-09
New ingame UI

SheridanR:
This PR does two things:

1. separate all the in-game hud code from main() into its own function called ingameHud() that runs each frame, this just made it logically easier to replace it
2. implement newIngameHud(), which creates a new hud layout supporting splitscreen for ingame usage (however, currently this hud is mostly non-functional)

This PR activates newIngameHud() by default, but the previous line just comments out ingameHud(), so it's painless to switch between them for testing and dev purposes

It's important to get this PR in sooner rather than later for the first bullet point. Rebasing this branch onto master today was a pain because the current ingame hud has received continuous updates while I've been working on this.


addictgamer:
- [ ] Looks like there's some build failures in CI, can you take a look at those?
For example:
```
/home/runner/work/Barony/Barony/src/ui/Frame.cpp:1341:24: error: non-const lvalue reference to type 'SDL_Rect' cannot bind to a temporary of type 'SDL_Rect'
                frame->setActualSize(SDL_Rect{0, 0, 200, 200});
```

addictgamer (commented):
Regarding the 3 new fonts: any in-repo licensing or attribution we need to include?

SheridanR:
> 
> 
> Regarding the 3 new fonts: any in-repo licensing or attribution we need to include?

ping mistersneak if he's got any word from the original creators, otherwise i'm inclined to treat them as if they were in the public domain.

SheridanR:
> 
> 
>     * [ ]  Looks like there's some build failures in CI, can you take a look at those?
>       For example:
> 
> 
> ```
> /home/runner/work/Barony/Barony/src/ui/Frame.cpp:1341:24: error: non-const lvalue reference to type 'SDL_Rect' cannot bind to a temporary of type 'SDL_Rect'
>                 frame->setActualSize(SDL_Rect{0, 0, 200, 200});
> ```

that's definitely pretty strange because it compiles without issues in visual studio. the fix would just be to assign a temp variable and pass a reference to that which is a bit clunky to me. I wonder if we can change CI to accept this or if I should change the function definition or something (I don't mind pass-by-value for SDL_Rect, frankly its probably faster)

addictgamer:
> that's definitely pretty strange because it compiles without issues in visual studio.

Ya, there's plenty of things that Ben wrote that worked in VS just fine, but Linux and Mac refused to compile them without minor changes...

src/ui/Frame.cpp
```diff
@@ -205,25 +238,25 @@ void Frame::draw(SDL_Rect _size, SDL_Rect _actualSize) {
...
+				drawDepressed(barRect.x, barRect.y, barRect.x + barRect.w, barRect.y + barRect.h);
 				break;
 			case BORDER_BEVEL_HIGH:
-				drawRect(&barRect, color, (Uint8)(color>>mainsurface->format->Ashift));
+				drawDepressed(barRect.x, barRect.y, barRect.x + barRect.w, barRect.y + barRect.h);
 				break;
 			case BORDER_BEVEL_LOW:
-				drawRect(&barRect, color, (Uint8)(color>>mainsurface->format->Ashift));
+				drawDepressed(barRect.x, barRect.y, barRect.x + barRect.w, barRect.y + barRect.h);
 				break;
```
addictgamer, line 256: It looks to me like it's doing the same thing in all 3 cases? Should we just remove the switch statement then? Or if we only want to do this if it's one of these 3 border types, one if statement checking if borderStyle == any of these 3?

addictgamer (changes requested):

src/ui/Frame.cpp
```diff
@@ -205,25 +238,25 @@ void Frame::draw(SDL_Rect _size, SDL_Rect _actualSize) {
...
 			}
 		} else {
-			drawRect(&barRect, color, (Uint8)(color>>mainsurface->format->Ashift));
+			drawDepressed(barRect.x, barRect.y, barRect.x + barRect.w, barRect.y + barRect.h);
```
addictgamer, line 259: Actually, it looks to me like even the else is doing the same thing, so we could entirely get rid of `if (border > 0) { ... } else ...` and instead just do `drawDepressed(barRect.x, barRect.y, barRect.x + barRect.w, barRect.y + barRect.h);`?

Unless my eyes accidentally scanned over a difference.

src/ui/Frame.cpp
```diff
@@ -1197,4 +1245,121 @@ void Frame::activateEntry(entry_t& entry) {
...
+
+	{
+		Button* bt = window->addButton("testButton2");
+		bt->setBorder(3);
+		bt->setSize(SDL_Rect{510, y, 240, 50});
+		bt->setText("Toggle button");
+		bt->setTooltip("Toggles on/off state");
+		bt->setStyle(Button::STYLE_TOGGLE);
+
+		y += 60;
```
addictgamer, line 1297: If we don't do this, will all of the buttons take up the same screen space?
Or is there a way to automatically figure out some of this for you? Like `y += bt->height()` or something, I guess...

src/ui/GameUI.cpp
```diff
@@ -0,0 +1,358 @@
...
+    static const char* bigfont = "fonts/pixelmix.ttf#18";
+    static const char* smallfont = "fonts/pixel_maz.ttf#32";
+
+    // big empty frame to serve as the root
+    if (playercount == 1) {
+        frame->setSize(SDL_Rect{0, 0, Frame::virtualScreenX, Frame::virtualScreenY});
+    } else if (playercount == 2) {
+        int y = (player % 2) * Frame::virtualScreenY / 2;
+        frame->setSize(SDL_Rect{0, y, Frame::virtualScreenX, Frame::virtualScreenY / 2});
+    } else if (playercount >= 3 || playercount <= 4) {
```
addictgamer, line 43: What about the 16 player beta branch? :)

addictgamer: I assume that playing a multiplayer game with 16 network but 1 local player works just fine here, right?

src/ui/GameUI.cpp
```diff
@@ -0,0 +1,358 @@
...
+        e->color = 0xffffffff;
+        e->text = "The dog hit smarts.";
+    }
+    {
+        auto e = chat->addEntry("chat", true);
+        e->color = 0xffff8800;
+        e->text = "Player 2: Chatlog text is the best, I swear my life on it.";
+    }
+
+    // other player statuses
```
addictgamer, line 83:
```suggestion
    // other players' statuses
```

src/ui/GameUI.cpp
```diff
@@ -0,0 +1,358 @@
...
+                "images/system/white.png",
+                "blue"
+            );
+            Field* text = mp_bar->addField("text", 64);
+            text->setSize(SDL_Rect{6, 3, 64, mp_bar->getSize().h - 4});
+            text->setText("MP");
+            text->setFont(smallfont);
+            text->setColor(0xffffffff);
+            text->setHJustify(Field::justify_t::LEFT);
+            text->setVJustify(Field::justify_t::CENTER);
```
addictgamer, line 327: Man, I wish we could have templates for these MP and HP bars. I feel like half this file is just duplicating this code for the player, their allies, and other players.

addictgamer:
> so it's painless to switch between them for testing and dev purposes

I think we've briefly talked about this, but we could make this a toggleable setting to switch between the legacy and new UI for our final product, right?

addictgamer (approved):
Everything else looks good to me!
Can't wait to play with the fully implemented new HUD.

(I didn't review game.cpp in any detail.)

src/ui/GameUI.cpp
WALLOFJUSTICE: Yeah multiplayer != SINGLE covers any server multiplayer configuration.

FWIW I've got the following helper functions I've been using to count players inside the player[] array
```
const bool Player::isLocalPlayer() const
{
	return ((splitscreen && bSplitscreen) || playernum == clientnum);
}
const bool Player::isLocalPlayerAlive() const
{
	return (isLocalPlayer() && entity && !client_disconnected[playernum]);
}
```

src/ui/GameUI.cpp
WALLOFJUSTICE: The intent eventually would be 2 image assets specifically for the player's HP/MP, then a different slimmer style for allies - ATM the wireframes are just rectangles so it's all very samey. Besides if we only write the code once here there's no point refactoring into a fancy catch-all draw function that'll eventually stop being suitable.

src/ui/Frame.cpp
SheridanR: The different cases are for different border styles which are not fully implemented.

src/ui/Frame.cpp
SheridanR: yeah, this is not declarative UI in the same way as HTML or something so each element still has to be placed manually (which has pros and cons). I do `y += 60` here so there's some padding between this element and the next one (I could have done `y += bt->getSize().h + 15`) not all the UI has to be defined exactly like this, that function is strictly a test case.

src/ui/GameUI.cpp
SheridanR: I'm totally open to the idea that some code here could be consolidated but these are small and unique elements that aren't gonna be replicated any more so I'm on the same line as ben here

SheridanR:
Thanks for the reviews, I fixed the issues and will merge as soon as the CI checks are clean

Issue #580, opened 2021-03-22
Broken textures with SDL 2.0.14


rapenne-s:
Hello,

I (and other people) are experiencing the same issue reported in #568 but the issue was closed so I reopen one.

I'm using Barony on OpenBSD and we now ship the latest SDL2 package with version 2.0.14 and we experience broken textures. We can't use older SDL2 .so files because of regular ABI changes that make them unusable (and it's not a great solution).

I'd be fine importing patches into our packaging to fix the issue if it's possible.

Is there a plan to make Barony compatible with newer SDL2 versions? That would be SO great for the future 👍

DanielGibson:
Sounds like an issue that should definitely be investigated - it could as well be a regression SDL2 that should be fixed there.

CyberOnion:
Running PopOS, I had to add the launch commands that was in issue 568 and afterwards the textures worked

DanielGibson:
Forcing the game to use an older SDL2 version is not a solution, it's a workaround at best.

namtsui:
cc @1bsyl

I bisected using the sdl2 hg repo and found the commit that caused this bug to show up. I get the missing textures as in: https://github.com/TurningWheel/Barony/issues/568

see: https://github.com/libsdl-org/SDL/commit/ebc12a2fd2bb692908884e7db6cc935941a591f2

```
dust2$ hg bisect --reset
dust2$ hg bisect --bad 14878  
dust2$ hg bisect --good 13608 
Testing changeset 14252:6c2bbcfc3940 (1213 changesets remaining, ~10 tests)
430 files updated, 0 files merged, 12 files removed, 0 files unresolved
dust2$ bisect --bad
ksh: bisect: not found
dust2$ hg bisect --bad
Testing changeset 13914:348d0896fc9e (606 changesets remaining, ~9 tests)
240 files updated, 0 files merged, 55 files removed, 0 files unresolved
dust2$ hg bisect --good
Testing changeset 14087:01d524d4574f (303 changesets remaining, ~8 tests)
107 files updated, 0 files merged, 0 files removed, 0 files unresolved
dust2$ hg bisect --bad  
Testing changeset 13997:738ede671341 (151 changesets remaining, ~7 tests)
54 files updated, 0 files merged, 12 files removed, 0 files unresolved
dust2$ hg bisect --good
Testing changeset 14035:e287d66e4210 (76 changesets remaining, ~6 tests)
[...]

PR #582, opened 2021-04-10
Fixed issue 580

1bsyl:
This should fixed the issue  https://github.com/TurningWheel/Barony/issues/580

- Don't use SDL_surface 'refcount', but 'userdata'
- Use imgref and not imgref + 1

### DIFF
--- a/src/draw.cpp
+++ b/src/draw.cpp
@@ -443,7 +443,7 @@ void drawImageRotatedAlpha( SDL_Surface* image, SDL_Rect* src, SDL_Rect* pos, re
 	}
 
 	// draw a textured quad
-	glBindTexture(GL_TEXTURE_2D, texid[image->refcount]);
+	glBindTexture(GL_TEXTURE_2D, texid[(long int)image->userdata]);
 	glColor4f(1, 1, 1, alpha / 255.1);
 	glBegin(GL_QUADS);
 	glTexCoord2f(1.0 * ((real_t)src->x / image->w), 1.0 * ((real_t)src->y / image->h));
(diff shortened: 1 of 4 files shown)


SheridanR (approved):
Looks good to me, well done.

perkinslr:
Commit ce9b618f89535b39a30d6ea0183eeb2f54b5d454 breaks the minimap (corrupted texture, eventual segfault) if I apply this patch to the 3.3.4 branch (last version available via Humble).  If it *doesn't* break later versions, that's fine, this comment will hopefully help anyone else in my position.  But it's also probably worth double-checking that it doesn't corrupt the minimap in the current master and 3.3.7 branch (just start a new game, if there's a problem, you'll get "static" on the bottom half of the minimap).

Without this commit, I haven't noticed any other issues, so I'm not sure what problem it is supposed to solve.  That said, the rest of this PR works as intended (and thanks for fixing the broken textures!)

1bsyl:
@perkinslr 
about https://github.com/TurningWheel/Barony/commit/ce9b618f89535b39a30d6ea0183eeb2f54b5d454

You can test the issue with latest (head) SDL_ttf, or simulate it by creating an intermediate surface with pitch != width 

Though, the patch may be wrong because:
- GL_UNPACK_ROW_LENGTH is by default in OpenGL whereas it's a only an extension for OpenGL ES2.
- this was assumed that bytes per pixel is 4 ...

To solve some potential issue, you can also manually pack the surface, if needed. 
like:
https://github.com/libsdl-org/SDL/blob/main/src/render/opengles2/SDL_render_gles2.c#L1503

PR #583, opened 2021-04-13
Initial STB TTF implementation

addictgamer:
* SDL_TTF still needs to be fully deprecated
 * There are a couple of integration issues that still need to be resolved, such as with gameui.cpp
 * We could also rework what features we make use of in STB/Fontstash, and what we build on top of them...(such as additional caching, multiline support, the whole concept of how we handle text/fonts, etc)

See also #584 for an alternate implementation (based off of this PR's branch)


SheridanR:
This implementation is a good start. We either need to upgrade FontStash to handle scaling of the result image + new lines, or we need to have FontStash render to a pixel buffer, and allow the Text class to handle scaling+coloring+caching of the image. I suspect the second method would be easier personally. As for newlines, that is a problem that can be tackled separately (seems exclusively like a FontStash/STB problem)

PR #584, opened 2021-04-13
 Second possible implementation for how we could do font stuff...

addictgamer:
Still doesn't address some of the basic potential issues, such as no scaling support and..."different" caching (this may not be a bad thing)

Main difference with #583 is it basically obsoletes the Text class, as the caching and scaling features are not currently supported...
That said,
* This still doesn't support scaling.
* If we want to scale:
   * Hack Fontstash to support scaling
   * Make Fontstash render to an SDL surface/create a bitmap and load an SDL surface out of that, and recommission the Text class.

(I also properly configured font alignment, that needs to get backported)

I personally prefer this implementation if it meets our needs, since the interface for drawing text is simpler.
Might not be possible depending on if we need scaling and how hard it is to make it play nicely with that.

Silly OpenGL question: If I want to scale and pretend I'm at a different resolution (even if the rest of the game really isn't), is changing the OpenGL viewport settings sufficient when we set that up before our text render calls?

### DIFF
--- a/src/ui/Button.cpp
+++ b/src/ui/Button.cpp
@@ -88,7 +88,7 @@ void Button::draw(SDL_Rect _size, SDL_Rect _actualSize) {
 			Font* _font = Font::get(font.c_str());
 			if (_font) {
 				int w, h;
-				_font->sizeText(text.c_str(), &w, &h);
+				_font->sizeText(text, &w, &h);
 				int x = (style != STYLE_DROPDOWN) ?
 					(size.w - w) / 2 :
 					5 + border;
@@ -117,6 +117,7 @@ void Button::draw(SDL_Rect _size, SDL_Rect _actualSize) {
 				scaledPos.y = pos.y * (float)yres / (float)Frame::virtualScreenY;
 				scaledPos.w = pos.w * (float)xres / (float)Frame::virtualScreenX;
 				scaledPos.h = pos.h * (float)yres / (float)Frame::virtualScreenY;
+				//TODO: If we want to go with this method, we have to recalculate the point size here...
 				_font->drawTextColor(text, scaledPos.x, scaledPos.y, textColor);
 			}
 		} else if (icon.c_str()) {
--- a/src/ui/Button.cpp
+++ b/src/ui/Button.cpp
@@ -85,10 +85,10 @@ void Button::draw(SDL_Rect _size, SDL_Rect _actualSize) {
 
 	if (style != STYLE_CHECKBOX || pressed) {
 		if (!text.empty()) {
-			Text* _text = Text::get(text.c_str(), font.c_str());
-			if (_text) {
-				int w = _text->getWidth(); //TODO: This won't work because text hasn't been drawn yet, so this value hasn't been cached yet...
-				int h = _text->getHeight();
+			Font* _font = Font::get(font.c_str());
+			if (_font) {
+				int w, h;
+				_font->sizeText(text.c_str(), &w, &h);
 				int x = (style != STYLE_DROPDOWN) ?
 					(size.w - w) / 2 :
 					5 + border;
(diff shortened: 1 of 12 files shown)


addictgamer:
> Silly OpenGL question: If I want to scale and pretend I'm at a different resolution (even if the rest of the game really isn't), is changing the OpenGL viewport settings sufficient when we set that up before our text render calls?

Eh, apparently it works but it's ugly. Maybe I should leave this problem to the graphics experts...

SheridanR:
The usefulness of Text comes from the notion that it represents a rendered image of some text using TTF (which has already been cached somewhere). I'm not too big a fan of changing to the font stash interface right now, particularly because it doesn't have built in support for scaling the resultant image... it is enough to me that we can render some text to a surface or buffer using a font, then let the Text class handle caching and drawing at different scales / colors.

src/ui/Button.cpp
```diff
@@ -117,7 +117,8 @@ void Button::draw(SDL_Rect _size, SDL_Rect _actualSize) {
 				scaledPos.y = pos.y * (float)yres / (float)Frame::virtualScreenY;
 				scaledPos.w = pos.w * (float)xres / (float)Frame::virtualScreenX;
 				scaledPos.h = pos.h * (float)yres / (float)Frame::virtualScreenY;
-				_text->drawColor(section, scaledPos, textColor);
+				//TODO: If we want to go with this method, we have to recalculate the point size here...
+				_font->drawTextColor(text, scaledPos.x, scaledPos.y, textColor);
```
addictgamer, line 121: If we don't scale a prerendered buffer, we need to modify the code at these points, whoever calls drawTextColor() needs to pass in the point size for their desired scaling...

addictgamer: Not sure on the exact implementation, but off the top of my head I might request to render my string at pointsize 10, take a measurement, then compute the maximum pointsize I can use to keep within the desired scaled dimensions.
Alternatively, we can just choose a standard size by UI designer's qualitive assessment and scale the UI element to fit it (if it's not big enough). Then if we ask to scale the UI from our standard resolution to whatever, 2x zoom, 4x zoom, 1/2x zoom, etc, we can just take a straightforward multiplication of the pointsize, maybe, to relatively scale with the element?

NOTE: Some languages may have interface elements that are much longer or shorter than English...how can we gracefully handle that?

src/ui/Font.cpp
```diff
@@ -29,8 +33,8 @@ Font::Font(const char* _name) {
 	} else {
 		path = name;
 	}
-	printf("Creating new font: %s\n", name.c_str());
 	fontstash = glfonsCreate(512, 512, FONS_ZERO_TOPLEFT);
+	//fontstash = glfonsCreate(512, 512, FONS_ZERO_BOTTOMLEFT);
```
addictgamer, line 37: Note to self:  before merge, need to go through this file and clean the junk comments up.

Also need to take one more pass over the function documentation and make sure I captured the bare minimums.

src/ui/Font.cpp
```diff
@@ -80,6 +135,65 @@ int Font::height() const {
 	}
 }
 
+int Font::textHeight(std::string str, int override_pointsize) const {
+	int numNewLines = std::count(str.begin(), str.end(), '\n');
+
+	return numNewLines * height(override_pointsize); //TODO: Do we need to account for a gap between every line??
```
addictgamer, line 141: I'm having some fonts overlap (but others don't), which is very strange and a bit unsettling. Need to look into this more...

src/ui/Font.hpp
```diff
@@ -16,26 +16,51 @@ class Font {
...
+	//! helper function for drawTextColor(), except operates on just one line.
+	void drawLineColor(std::string str, int x, int y, const Uint32& color, int override_pointsize);
+
 public:
 	//! built-in font
 	static const char* defaultFont;
 
+	//! tab-width in spaces
+	static const int TAB_WIDTH = 4;
+	//! The following is basically a prerendered tab replacement...must be maintained to match tab_width. //TODO: Why even have TAB_WIDTH, then?
```
addictgamer, line 31: The TODO here is a lie, we need TAB_WIDTH instead if we want true tab behavior...need to fix this, tab currently just always inserts `TAB_WIDTH` number of spaces, which is not the correct behavior. It should align up to the next multiple of `TAB_WIDTH`

PR #588, opened 2021-06-04
Fixed compilation on Linux

addictgamer:


addictgamer:
Huh. I didn't spot any failure messages...

PR #592, opened 2021-07-25
Fix UI scaling

SheridanR:
UI is now rendered to its own framebuffer (render-to-texture) then drawn on a fullscreen quad to fix seams between ui elements at non-720p resolutions.

Also unlocks Frame::process() from the original tickrate which was 50.


src/init.cpp
```diff
@@ -399,10 +417,13 @@ int initApp(char const * const title, int fullscreen)
 	}
 
 	// create player classes
+	// TODO/FIXME: why isn't this in initGame? why is it in init.cpp?
```
WALLOFJUSTICE, line 420: This was to not crash on frames checking for the mouse in the loading screen (caught this in debug)

SheridanR: I see, this PR changes the UI code so it reads mousex/mousey etc variables directly when EDITOR is defined (so it doesn't crash). ergo this can probably go in initGame() now

WALLOFJUSTICE: I think it was crashing the main game, not editor

src/ui/GameUI.cpp
```diff
@@ -715,7 +715,7 @@ void createInventoryTooltipFrame(const int player)
 		auto tooltipTextField = tooltipFrame->addField("inventory mouse tooltip header", 1024);
 		tooltipTextField->setText("Nothing");
 		tooltipTextField->setSize(SDL_Rect{ 0, 0, 0, 0 });
-		tooltipTextField->setFont("fonts/pixelmix.ttf#14");
+		tooltipTextField->setFont("fonts/pixelmix.ttf#16");
```
WALLOFJUSTICE, line 718: Could you revert these font size changes for my HUD stuff - I'll need to redo the spacing/layout and want the old sizes for reference.

PR #595, opened 2021-11-03
finish-settings

SheridanR:
Quick change-list:
- Implemented pause menu
- Added quit game prompt, disconnect prompt, restart game, end-life, etc.
- 100% Settings menu (Changing video mode works, and Minimap, Messages, and input Binding menus are all functional)
- UI engine changes: draw() functions are now const, and Text can now have a unique outline color

### DIFF
--- a/src/book.cpp
+++ b/src/book.cpp
@@ -620,7 +620,8 @@ void BookParser_t::createBook(std::string filename)
 		firstIteration = false;
 		pageText += token;
 		tmpField->setText(pageText.c_str());
-		if ( auto getText = Text::get(tmpField->getText(), tmpField->getFont()) )
+		if ( auto getText = Text::get(tmpField->getText(), tmpField->getFont(),
+			makeColor(255, 255, 255, 255), makeColor(0, 0, 0, 255)) )
 		{
 			int textHeight = getText->getHeight();
 			if ( textHeight > tmpField->getSize().h )
@@ -694,7 +695,8 @@ void BookParser_t::createBook(std::string filename)
 	//		}
 	//	}
 	//	tmpField.setText(pageText.c_str());
-	//	if ( auto getText = Text::get(tmpField.getText(), tmpField.getFont()) )
+	//	if ( auto getText = Text::get(tmpField.getText(), tmpField.getFont(),
+	//		makeColor(255, 255, 255, 255), makeColor(0, 0, 0, 255)) )
 	//	{
 	//		if ( getText->getHeight() > tmpField.getSize().h - tmpField.getSize().y )
 	//		{
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -5113,7 +5113,8 @@ int main(int argc, char** argv)
 						inputs.controllerClearInput(clientnum, INJOY_MENU_NEXT);
 						inputs.controllerClearInput(clientnum, INJOY_MENU_CANCEL);
 						fadealpha = 255;
-#if (!defined STEAMWORKS && !defined USE_EOS && !defined NINTENDO)
+						// Yeah we're just not going to do the "Please don't pirate us" message anymore
+#if (0)
 						introstage = 0;
 						fadeout = false;
 						fadefinished = false;
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -2850,6 +2850,7 @@ void handleEvents(void)
 		inputs.updateAllMouse();
 	}
 
+	Input::lastInputOfAnyKind = "";
 	for (auto& input : Input::inputs) {
 		input.updateReleasedBindings();
 		input.update();
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -3417,16 +3417,6 @@ void handleEvents(void)
 						printlog("critical error! Attempting to abort safely...\n");
 						mainloop = 0;
 					}
-					if (zbuffer != NULL)
-					{
-						free(zbuffer);
-					}
-					zbuffer = (real_t*)malloc(sizeof(real_t) * xres * yres);
-					if (clickmap != NULL)
-					{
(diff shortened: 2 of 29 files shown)


src/book.cpp
```diff
@@ -620,7 +620,8 @@ void BookParser_t::createBook(std::string filename)
 		firstIteration = false;
 		pageText += token;
 		tmpField->setText(pageText.c_str());
-		if ( auto getText = Text::get(tmpField->getText(), tmpField->getFont()) )
+		if ( auto getText = Text::get(tmpField->getText(), tmpField->getFont(),
```
addictgamer, line 623: Note to self: this is just a `Text::get()` update

src/menu.hpp
```diff
@@ -283,6 +283,8 @@ extern Sint32 oldXres;
 extern Sint32 oldYres;
 extern button_t* revertResolutionButton;
 
+typedef std::tuple<int, int> resolution;
```
addictgamer, line 286: I like it! Named datatypes sure make the code so much more reasonable! ("Mommy, why is he passing in _standard tuple int int_")

src/ui/GameUI.cpp
```diff
@@ -22,7 +22,7 @@
 
 #include <assert.h>
 
-bool newui = false;
+bool newui = true;
```
addictgamer, line 25: Is it time to go official? :)

src/ui/Field.hpp
```diff
@@ -65,10 +65,15 @@ class Field : public Widget {
 	//! gets the physical screen-space x/y (not relative to current parent - but to the absolute root)
 	SDL_Rect getAbsoluteSize() const;
 
+	//! TODO comment pls
```
addictgamer, line 68: yes pls

src/init_game.cpp
```diff
@@ -282,7 +282,7 @@ int initGame()
 			}
 		}
 		FileIO::close(fp);
-		bookParser_t.createBooks(false);
```
addictgamer, line 285: What was this doing? Why are we disabling it?

addictgamer: (I would love to see a code review IDE that lets me explore the code more contextually...or I guess I can check out this branch next time I pick up this review)

SheridanR: Good question, I had to disable this because it was trying to render and cache text from another thread which is not safe. Books are now "compiled" and cached offline and stored in a json file, so this line is not needed to run the game in release mode as long as you have that data file. You'd have to ask @WALLOFJUSTICE how this data file is created now (I think he just put it outside of this init thread).

src/ui/Frame.cpp
```diff
@@ -664,79 +668,93 @@ Frame::result_t Frame::process(SDL_Rect _size, SDL_Rect _actualSize, const std::
...
-					mousestatus[SDL_BUTTON_WHEELDOWN] = 0;
-					this->actualSize.x += std::min(entrySize * 4, size.w);
-					usable = result.usable = false;
-				} else if (mousestatus[SDL_BUTTON_WHEELUP]) {
-					mousestatus[SDL_BUTTON_WHEELUP] = 0;
-					this->actualSize.x -= std::min(entrySize * 4, size.w);
-					usable = result.usable = false;
-				}
-			}
+	if (parent != nullptr && !hollow && rectContainsPoint(fullSize, omousex, omousey) && usable) {
```
WALLOFJUSTICE, line 718: `allowScrolling && allowScrollBinds` was removed, this consumes the mouse wheel when any frame is open and doesn't check if scrolling is needed

src/game.cpp
```diff
@@ -5122,7 +5116,8 @@ int main(int argc, char** argv)
 						inputs.controllerClearInput(clientnum, INJOY_MENU_NEXT);
 						inputs.controllerClearInput(clientnum, INJOY_MENU_CANCEL);
 						fadealpha = 255;
-#if (!defined STEAMWORKS && !defined USE_EOS && !defined NINTENDO)
+						// Yeah we're just not going to do the "Please don't pirate us" message anymore
```
addictgamer, line 5119: Resolves #23 from the NX issue tracker

src/game.cpp
```diff
@@ -5758,33 +5753,21 @@ int main(int argc, char** argv)
 
 				doFrames();
 
-				if ( !gamePaused )
+				if ( /*newui*/ 0 )
```
addictgamer, line 5756: umm...

addictgamer: Is there a better way to do this? I had to do a double take when I read that line the first time

src/init_game.cpp
addictgamer: @WALLOFJUSTICE will a new, clean install of the game still automatically generate this cached file on first startup?

src/game.cpp
WALLOFJUSTICE: I've removed it in my branch, deprecated anyway should always just be ingameHud()

src/init_game.cpp
WALLOFJUSTICE: Yes I had moved this below the asynchronous task, after loading completes. Takes 10 seconds or so. For some reason debug builds were OK on my machine,  but other machines/release mode would crash as the text cache changes data to render out the books. 

Thinking now, maybe could make temporary text cache just for this purpose that won't touch the Text() hashes used for the loading screen and it could solve the issue

https://github.com/WALLOFJUSTICE/Barony/commit/545b19f51e5c4113ee70b67140393dd8f0311d3e

src/game.cpp
addictgamer: Ah ok good to know, thanks (haven't had a chance to look over your branch yet)

PR #598, opened 2021-11-22
Continue UI

WALLOFJUSTICE:


src/ui/Text.cpp
```diff
@@ -136,18 +136,14 @@ void Text::render() {
 		glGenTextures(1, &texid);
 	}
 
-	width = 0;
+	width = surf->w;
+	height = surf->h;
+
+	// Fields break multi-lines anyway, and we're not using TTF_RenderUTF8_Blended_Wrapped()
```
WALLOFJUSTICE, line 142: Regressions of changing this:
- Any mention of Text::get()->getHeight() will only return the height of 1 line. To accurately size text you need to multiply by number of lines. Breaks message log/book compiling ATM which didn't assume necessary to multiply by text lines. 
 
Is this 100% what we want? The stored height of the Field is not accurate to the text contained within, so either store the accurate height at render() time, or add getHeight() * getNumTextLines() everywhere to calculate the true height of the field.

src/interface/clickdescription.cpp
```diff
@@ -365,6 +365,32 @@ void clickDescription(int player, Entity* entity)
...
+						(entity->sprite == 620))
+					{
+						messagePlayer(player, language[272]);
+					}
+				}
+				else if ( entity->behavior == &actFloorDecoration )
+				{
+					if ( entity->sprite == 991 )
+					{
+						messagePlayer(player, language[4048]);
```
WALLOFJUSTICE, line 391: Language file is not updated in your branch, 4048 is already taken. 

I suggest language[4073], my latest lang file: https://github.com/WALLOFJUSTICE/Barony/blob/dev-20-merge-main-menu/lang/en.txt

src/ui/Text.cpp
SheridanR: Text::get()->getHeight() should in fact return the height of the rendered text image, so if there are line-breaks in the text you will see that reflected in the height of the image. However, if you render an empty string, it is possible the height of the rendered image would be 0, so blank lines of text in a book may be responsible for the errors here (though in truth, I don't really know why this is happening or what to expect in that case).

I think the proper solution in most cases where you want to pre-determine the pixel size of multiple lines of text, is actually to use Font::get()->height() * getNumTextLines(). This should give you a pretty accurate estimate and also avoid complications that arise from the use of glyphs that are abnormally large or small, or exceed the usual boundaries of a single character.

SheridanR: Confirmed createBooks() is broken btw, I'll fix it and commit the change to this branch.

SheridanR: Fixed

PR #600, opened 2021-12-02
Fixed a crash on game restart after death.

addictgamer:
Looks like an issue from the temporary interaction of the old menu code and the new menu code.

### DIFF
--- a/src/menu.cpp
+++ b/src/menu.cpp
@@ -50,6 +50,7 @@
 
 #include "ui/Text.hpp"
 #include "ui/Font.hpp"
+#include "ui/MainMenu.hpp"
 
 #ifdef STEAMWORKS
 //Helper func. //TODO: Bugger.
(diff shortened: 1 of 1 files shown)


src/menu.cpp
```diff
@@ -13389,6 +13390,7 @@ void buttonStartSingleplayer(button_t* my)
 	numplayers = 0;
 	introstage = 3;
 	fadeout = true;
+	MainMenu::beginFade(MainMenu::FadeDestination::GameStart);
```
addictgamer, line 13393: Not sure how safe it is to interleave new menu code in the old menu code. I assume all of this old menu code will eventually be superseded by the new menu code, and situations like these are just temporary as we continue in-dev.

WALLOFJUSTICE: Correct the old death prompt (purple panel/buttons) will need to be upgraded and then will use the correct new functions

PR #602, opened 2021-12-09
+ Set up a CI for Windows builds.

addictgamer:
Our build config files are terrible and potato king should feel ashamed of himself.

I want to do a little bit more cleanup before we merge this. I probably botched some random stuff to get this working.


.github/workflows/build-windows_fmod_steam.yml
```diff
@@ -0,0 +1,69 @@
...
+    branches: [ master, develop ]
+  pull_request:
+    branches: [ master, develop ]
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: windows-latest
```
addictgamer, line 18: I think we'll want to hardcode this to Windows-10 in at least 1 CI lane

.github/workflows/build-windows_fmod_steam.yml
```diff
@@ -0,0 +1,69 @@
...
+        unzip dependencies_windows_x64.zip -d dependencies
+      shell: bash
+
+    # Actually does the build...
+    - name: Generate Visual Studio project via CMake
+      run: |
+        echo "Generating Visual Studio project"
+        mkdir build
+        cd build
+        cmake .. -DSTEAMWORKS_ENABLED=1 -DEOS_ENABLED=0 -DFMOD_ENABLED=ON -DBARONY_WIN32_LIBRARIES="$(pwd)/../dependencies/gamelibs/"
```
addictgamer, line 47: Need to add one for building barony.exe with EOS.

.github/workflows/build-windows_fmod_steam.yml
```diff
@@ -0,0 +1,69 @@
...
+      shell: bash
+  
+    - name: Compile
+      # Add additional options to the MSBuild command line here (like platform or verbosity level).
+      # See https://docs.microsoft.com/visualstudio/msbuild/msbuild-command-line-reference
+      run: msbuild /m build/Barony.sln /property:Configuration=Release
```
addictgamer, line 53: Need to check if this builds 32bit or 64bit...I think the dependencies zip were 64 bit libraries, and the build wasn't complaining, so we're probably ok? Will double check.

PR #603, opened 2021-12-09
Added a CI build for Windows

addictgamer:
Our build config files are terrible and potato king should feel ashamed of himself.

I want to do a little bit more cleanup before we merge this. I probably botched some random stuff to get this working.

(Reopening under this repo since forks don't have access to secrets)


addictgamer:
```
> file barony.exe
barony.exe: PE32+ executable (GUI) x86-64, for MS Windows
```

Woot! Looks like it is 64 bit.
Will need to boot into Windows sometime to actually test it.

cmake/Modules/FindSDL2_ttf.cmake
```diff
@@ -33,6 +33,7 @@ find_path(SDL2_TTF_INCLUDE_DIR SDL_ttf.h
   HINTS
     ENV SDLTTFDIR
     ENV SDLDIR
+	${SDL2DIR}
```
addictgamer, line 36: Oof, indentation mismatch that wasn't visible in my editor. :|

PR #604, opened 2021-12-13
Latest

WALLOFJUSTICE:


SheridanR:
Text draw functions should be const, render() gets called in the constructor now so there's no reason to call it in either of the draw() functions. Other than that looks fine to me.

PR #612, opened 2022-01-08
Bindings/gamepad stuffs

WALLOFJUSTICE:


src/input.hpp
```diff
@@ -162,15 +183,28 @@ class Input {
 	std::string getGlyphPathForInput(binding_t binding) const;
 	std::string getGlyphPathForInput(const char* binding) const;
 
+	static float getJoystickRebindingDeadzone() { return rebinding_deadzone; }
+
 	//! list of connected input devices
 	static std::string lastInputOfAnyKind;
+	static int Input::waitingToBindControllerForPlayer;
```
SheridanR, line 190: GCC complains about an extra `Input::` qualifier on this line

PR #616, opened 2022-01-25
More stuff

SheridanR:
Made some more stuff. Check commits


src/interface/interface.cpp
```diff
@@ -874,298 +846,6 @@ int saveConfig(char const * const _filename)
...
-	if ( disableMultithreadedSteamNetworking )
-	{
-		fp->printf("/disablenetworkmultithreading\n");
-	}
-	if ( disableFPSLimitOnNetworkMessages )
-	{
-		fp->printf("/disablenetcodefpslimit\n");
-	}
-#ifdef USE_EOS
-	if ( LobbyHandler.crossplayEnabled )
```
WALLOFJUSTICE, line 1140: Needs a new equivalent saved into the config file before we delete

src/interface/interface.cpp
```diff
@@ -874,298 +846,6 @@ int saveConfig(char const * const _filename)
...
-	{
-		fp->printf("/crossplay\n");
-	}
-#endif // USE_EOS
-
-	if ( !gamemods_mountedFilepaths.empty() )
-	{
-		std::vector<std::pair<std::string, std::string>>::iterator it;
```
WALLOFJUSTICE, line 1148: Needs a new equivalent saved into the config file before we delete

src/magic/magic.hpp
```diff
@@ -73,6 +73,8 @@ static const int SPELL_DASH = 51;
 static const int SPELL_SELF_POLYMORPH = 52;
 static const int SPELL_9 = 53;
 static const int SPELL_10 = 54;
+static const int SPELL_CRAB_FORM = 55;
```
WALLOFJUSTICE, line 76: Would be better to set these to higher numbers we'll never reach e.g 10000, 10001. New spells will require these changing as NUM_SPELLS increases as these aren't real spells. Though does this need to be sequential to fetch the item's icon?

WALLOFJUSTICE (changes requested):
Merge conflicts:

- [x] Language file needs to be manually merged, no conflicting language lines

- [x] GameUI.cpp - keep my changes for the selectors, but re-add your block for the console command window

- [ ] Widget.hpp - manually merge our 2 changed lines

src/magic/magic.hpp
SheridanR: these are strictly used to get the item icon index. so they need to be sequential.

SheridanR: these are not real spells - just numbers with a name

WALLOFJUSTICE: Yep just use spell 9 and 10 indices and we'll skip em for next spells

src/game.cpp
```diff
@@ -5895,7 +5906,11 @@ int main(int argc, char** argv)
 			// fps counter
 			if ( showfps )
 			{
-				printTextFormatted(font8x8_bmp, 8, 8, "fps = %3.1f", fps);
+			    char buf[16];
```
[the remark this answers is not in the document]
SheridanR: done

src/interface/interface.cpp
SheridanR: done

src/interface/interface.cpp
SheridanR: done

src/magic/magic.hpp
SheridanR: done

PR #623, opened 2022-02-18
Chests

WALLOFJUSTICE:


src/collision.cpp
```diff
@@ -99,17 +99,18 @@ Entity* entityClicked(bool* clickedOnGUI, bool clickCheckOverride, int player, E
...
-				&& my > getChestGUIStartY(player)
-				&& my < getChestGUIStartY(player) + inventoryChest_bmp->h )
-			{
-				if ( clickedOnGUI )
-				{
-					*clickedOnGUI = true;
-				}
-				return NULL;    //Click falls inside the chest inventory GUI.
-			}
+			// TODO UI: REPLACE
```
SheridanR, line 102: Does this still need to be handled? Is it only the buttons below the window that need this, or the whole window?

src/entity.cpp
```diff
@@ -3699,7 +3699,23 @@ void Entity::handleEffects(Stat* myStats)
 				{
 					if ( myStats->shield->count > 1 )
 					{
-						newItem(myStats->shield->type, myStats->shield->status, myStats->shield->beatitude, myStats->shield->count - 1, myStats->shield->appearance, myStats->shield->identified, &myStats->inventory);
+						Uint32 newAppearance = rand();
+						int tries = 100;
+						while ( tries > 0 && newAppearance == myStats->shield->appearance )
+						{
+							--tries;
```
SheridanR, line 3706: this decrements `tries` but doesn't `newAppearance = rand();` also maybe it would be better to just:
```
newAppearance = rand();
if (newAppearance == myStats->shield->appearance)
    newAppearance++;
```

?

src/input.cpp
```diff
@@ -286,7 +289,16 @@ void Input::refresh() {
 		prefix.append("Pad");
 		prefix.append(std::to_string(player));
 		for (auto& binding : getGamepadBindings()) {
-		    bind(binding.first.c_str(), (prefix + binding.second).c_str());
+			if ( binding.second == MainMenu::hiddenBinding )
```
SheridanR, line 292: wondering if there isn't a more graceful way to do this

src/items.hpp
```diff
@@ -621,7 +622,26 @@ void clientSendEquipUpdateToServer(EquipItemSendToServerSlot slot, EquipItemResu
 	ItemType type, Status status, Sint16 beatitude, int count, Uint32 appearance, bool identified);
 void clientUnequipSlotAndUpdateServer(const int player, EquipItemSendToServerSlot slot, Item* item);
 EquipItemResult equipItem(Item* item, Item** slot, int player, bool checkInventorySpaceForPaperDoll);
-Item* itemPickup(int player, Item* item);
+enum ItemStackResults : int
```
SheridanR, line 625: I'm beginning to understand why you needed a spreadsheet for this

src/main.hpp
```diff
@@ -623,7 +623,7 @@ extern int minotaurlevel;
 #define SPLITSCREEN 6
 
 // language stuff
-#define NUMLANGENTRIES 4100
+#define NUMLANGENTRIES 4200
```
SheridanR, line 626: Everytime I think about the lang stuff I feel like it needs an overhaul. Should come after the open beta I think

src/player.cpp
```diff
@@ -487,36 +527,14 @@ void GameController::handleAnalog(int player)
...
-
 int GameController::getRightXMove() // with sensitivity
 {
 	if (!isActive())
 	{
 		return 0;
 	}
 	int x = getRawRightXMove();
-	x *= gamepad_rightx_sensitivity;
+	x *= getGamepadRightXSensitivity(0);
```
SheridanR, line 537: should this not take a proper player index?

src/player.cpp
```diff
@@ -527,7 +545,7 @@ int GameController::getRightYMove() // with sensitivity
 		return 0;
 	}
 	int y = getRawRightYMove();
-	y *= gamepad_righty_sensitivity;
+	y *= getGamepadRightYSensitivity(0);
```
SheridanR, line 548: should this not take a proper player index?

src/ui/Frame.cpp
```diff
@@ -1772,240 +1772,236 @@ void createTestUI() {
 }
 
 // sample function - would need to cache the blitted images somewhere for real-time use.
```
SheridanR, line 1774: got an idea of the performance penalty here? could put this functionality right in Image class - eg add some characters to the filename to cache an image with an outline. like `Image::get("myimage.png#0xffffffff#2")`

SheridanR (approved):
Looks solid, skimmed GameUI.cpp and checked everything else pretty closely. Left comments for potential things to change at your prerogative

src/collision.cpp
WALLOFJUSTICE: Assuming that the click detection on frames solves most of the detection, floating widgets like Buttons may/may not need some extras added. collision.cpp/clickdescription.cpp are the 2 spots that need this deep dive

src/entity.cpp
WALLOFJUSTICE: I wasn't happy with this solution, reverted the appearance change. This pattern appears a lot across the code and I'd rather look at it at some other stage

src/player.cpp
WALLOFJUSTICE: Done - splitscreen will need this sensitivity unique for each player since everyone just uses the global sensitivity

src/ui/Frame.cpp
WALLOFJUSTICE: If I made everything with an outline, the game drops from 140FPS shootmode/100FPS !shootmode to 100/60FPS in debug. Importantly the outline SDL surface is only created once and stored within the image, though each time it draws it needs to texture->bind() texture->load() like the minimap. Can refine it further if we ever have a use for it

PR #624, opened 2022-02-18
Network Lobbies

SheridanR:


src/ui/MainMenu.cpp
```diff
@@ -3985,25 +4166,1324 @@ namespace MainMenu {
...
+	    directConnect = false;
+	    if (LobbyHandler.getHostingType() == LobbyHandler_t::LobbyServiceType::LOBBY_STEAM) {
+#ifdef STEAMWORKS
+		    for ( int c = 0; c < MAXPLAYERS; c++ ) {
+			    if ( steamIDRemote[c] ) {
+				    cpp_Free_CSteamID(steamIDRemote[c]);
+				    steamIDRemote[c] = NULL;
+			    }
+		    }
+		    ::currentLobbyType = k_ELobbyTypePrivate;
```
WALLOFJUSTICE, line 5245: Invalid assignment here - two different enums. I'm sure you want to CreateLobby() using k_ELobbyTypePrivate but ::currentLobbyType should be something like `LobbyType::LobbyOnline`?

SheridanR: ::currentLobbyType refers to the definition in steam.hpp, so this compiles fine for me

WALLOFJUSTICE: My b, still was using code from Friday testing out steamworks builds now

src/ui/MainMenu.cpp
```diff
@@ -3985,25 +4166,1324 @@ namespace MainMenu {
...
+				}
+
+				continue;
+			}
+
+			// network scan
+			else if (packetId == 'SCAN') {
+			    if (directConnect) {
+			        char hostname[256] = { '\0' };
+			        (void)gethostname(hostname, sizeof(hostname));
```
WALLOFJUSTICE, line 4616: Note I'm getting missing definition here for `gethostname`

src/ui/MainMenu.cpp
```diff
@@ -6054,26 +7541,118 @@ namespace MainMenu {
...
+	        // set unique game key
+	        uniqueGameKey = prng_get_uint();
+	        if (!uniqueGameKey) {
+		        uniqueGameKey++;
+	        }
+
+	        // send start signal to each player
+	        if (multiplayer == SERVER) {
+	            for (int c = 1; c < MAXPLAYERS; c++) {
+		            if (client_disconnected[c] || players[c]->isLocalPlayer()) {
```
WALLOFJUSTICE, line 7642: Can't use `players[c]->isLocalPlayer()` here as `intro` is true and every client returns true on the lobby screen

SheridanR: fair enough, will just change it to `c == clientnum`

src/ui/MainMenu.cpp
SheridanR: ah yes, forgot this is a POSIX function. Fun times.

PR #639, opened 2022-03-08
Hook up cutscenes

SheridanR:


src/actladder.cpp
```diff
@@ -488,19 +499,61 @@ void actWinningPortal(Entity* my)
 						}
 						strcpy((char*)net_packet->data, "WING");
 						net_packet->data[4] = victory;
+						net_packet->data[5] = cutscene;
 						net_packet->address.host = net_clients[c - 1].host;
 						net_packet->address.port = net_clients[c - 1].port;
-						net_packet->len = 8;
+						net_packet->len = 5;
```
WALLOFJUSTICE, line 505: should be `net_packet->len = 6;` due to cutscene var

src/actladder.cpp
```diff
@@ -488,19 +499,61 @@ void actWinningPortal(Entity* my)
...
-				introstage = 5; // prepares win game sequence
-				fadeout = true;
-				if ( !intro )
-				{
-					pauseGame(2, false);
-				}
+
+	            if (cutscene == 1) { // classic herx ending
+	                switch (stats[clientnum]->playerRace) {
+	                default:
```
WALLOFJUSTICE, line 512: TIL the concept of putting `default:` anywhere but the bottom 😄

src/actladder.cpp
```diff
@@ -786,56 +857,81 @@ void Entity::actMidGamePortal()
...
+	                case RACE_SKELETON:
+	                case RACE_VAMPIRE:
+	                case RACE_SUCCUBUS:
+	                case RACE_INCUBUS:
+	                    MainMenu::beginFade(MainMenu::FadeDestination::BaphometMidpointEvil);
+	                    break;
+	                }
+	            }
+
+                movie = true;
```
WALLOFJUSTICE, line 933: Reminder to check Entity::isMobile to make sure monsters don't move around in cutscenes

src/entity.cpp
```diff
@@ -5766,13 +5766,9 @@ bool Entity::isMobile()
...
 
-	if ( behavior == &actMonster && 
-		(introstage == 9
-		|| introstage == 11 + MOVIE_MIDGAME_BAPHOMET_HUMAN_AUTOMATON
-		|| introstage == 11 + MOVIE_MIDGAME_BAPHOMET_MONSTERS
-		|| introstage == 11 + MOVIE_MIDGAME_HERX_MONSTERS) )
+	if ( behavior == &actMonster && MainMenu::isCutsceneActive() )
 	{
-		return false; // mid-game crawls.
+		return false;
```
WALLOFJUSTICE, line 5771: Verify this is working correctly? I.e don't get slaughtered by demons in hell while watching cutscene in network lobby

src/net.cpp
```diff
@@ -351,37 +351,20 @@ bool messagePlayerColor(int player, Uint32 type, Uint32 color, char const * cons
...
-		    }
-		    if ( !disable_messages )
-		    {
-			    players[player]->messageZone.addMessage(color, buf);
-		    }
-		    ptr += c + 1;
-		} while (!end);
+	    printlog("%s\n", str);
+	    newString(&messages, color, str);
+	    while ( list_Size(&messages) > MESSAGE_LIST_SIZE_CAP )
```
WALLOFJUSTICE, line 363: Do messages process / expire for disconnected players? Like reverting using `/splitscreen` to singleplayer. Used to get backed up

src/entity.cpp
SheridanR: It works fine but I added another if check to actMonster(). it returns immediately `if (movie || MainMenu::isCutsceneActive())`

src/net.cpp
SheridanR: I don't think I fixed that issue. OTOH that's not possible in a normal game.

src/actladder.cpp
SheridanR: Thanks, I was afraid I messed up one of those.

PR #642, opened 2022-03-22
Leaderboards

SheridanR:


src/ui/MainMenu.cpp
```diff
@@ -9341,197 +10444,45 @@ namespace MainMenu {
...
-		    "join_image"
-		);
-
-		auto tooltip = window->addField("tooltip", 1024);
-		tooltip->setSize(SDL_Rect{106, 398, 300, 48});
-		tooltip->setFont(smallfont_no_outline);
-		tooltip->setColor(makeColor(91, 76, 50, 255));
-		tooltip->setJustify(Field::justify_t::CENTER);
-		tooltip->setText("Help text goes here.");
+	    if (skipintro) {
```
WALLOFJUSTICE, line 10447: There's a bool FirstTimePrompt.showFirstTimePrompt for this, which gets written to disk in the tutorial_scores.json "first_time_prompt" with the Tutorial.writeToDocument() on button click

Recommend adding it instead of skipintro - could close the game prematurely at title screen (couldn't press 'A' or something on controller) and skipintro would be false next boot?

src/list.cpp
```diff
@@ -457,14 +457,11 @@ Uint32 list_Index(node_t* node)
 
 node_t* list_Node(list_t* list, int index)
 {
-	if ( index < 0 || index >= list_Size(list) )
-	{
+	if (index < 0) {
```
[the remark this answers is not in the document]
SheridanR: The old method was just inefficient, this function still returns nullptr if index is beyond the list size.

the problem was that list_Size() walks the entire list to return the size which is redundant because we're already walking the list.

After the for statement is completed, `node` will be nullptr because `node->next` will be nullptr on the last node. So after for () completes, node will always be either the correct node or nullptr if the node was not found.

src/ui/MainMenu.cpp
SheridanR: Fixed

PR #647, opened 2022-05-03
Obituaries

SheridanR:


SheridanR:
server needs to transmit obituaries to client before they open gameover window

PR #654, opened 2022-05-20
Skr/log and minimap

SheridanR:


src/net.cpp
```diff
@@ -3677,27 +3661,61 @@ void clientHandlePacket()
...
+		loading = false;
+	    int result = loading_task.get();
 
 		// (special) unlock temple achievement
 		if ( secretlevel && currentlevel == 8 )
 		{
 			steamAchievement("BARONY_ACH_TRICKS_AND_TRAPS");
 		}
 
+		saveGame();
```
WALLOFJUSTICE, line 3718: Should savegame() be before the async task gets started? Perhaps there could be potential crashes in loadMap if you have map data from a mod

src/init.cpp
```diff
@@ -2650,52 +2612,15 @@ bool initVideo()
...
-#ifndef EDITOR
 	Frame::fboDestroy();
-#endif
-
-	// delete vertex data
-	if ( !disablevbos )
-	{
-		for ( c = 0; c < nummodels; c++ )
-		{
-			SDL_glDeleteBuffers(1, &polymodels[c].vbo);
```
WALLOFJUSTICE, line 2677: Did any of these polymodels buffers need to be deleted?

WALLOFJUSTICE (commented):
Commented

src/init.cpp
SheridanR: They were deleted previously because the gl context was being destroyed leaving them invalid.

They no longer need to be deleted.

src/net.cpp
SheridanR: Good idea.

PR #655, opened 2022-05-26
Test

WALLOFJUSTICE:


SheridanR (approved):
Shopkeepers should probably just be neutral towards humans, dummy bots, etc. instead of having special case code for incitement

PR #658, opened 2022-06-07
Skr/lobby work

SheridanR:


src/net.cpp
```diff
@@ -3655,8 +3655,8 @@ void clientHandlePacket()
 		{
 			minimapPings[i].clear(); // clear minimap pings
 			enemyHPDamageBarHandler[i].HPBars.clear();
+		    players[c]->camera().globalLightModifierActive = GLOBAL_LIGHT_MODIFIER_STOPPED;
```
WALLOFJUSTICE, line 3658: [c] should be [i]

src/opengl.cpp
```diff
@@ -1935,7 +1940,7 @@ void glDrawWorld(view_t* camera, int mode)
 		glEnable(GL_BLEND);
 
 		// first (higher) sky layer
-		glColor4f(1.f, 1.f, 1.f, .5);
+		glColor4f(1.f, 1.f, 1.f, getLightAtModifier);
```
WALLOFJUSTICE, line 1943: Fine to remove the .5?

SheridanR: yeah i just decided i didn't like the top layer of sky being transparent ... the two layers combined weren't opaque

PR #672, opened 2022-09-20
Minotaur/Scroll fx guis

WALLOFJUSTICE:


src/mod_tools.cpp
```diff
@@ -896,10 +896,11 @@ void ItemTooltips_t::readTooltipsFromFile()
 	{
 		printlog("[JSON]: Error: file size is too large to fit in buffer! %s", inputPath.c_str());
 		FileIO::close(fp);
+		return;
 	}
 
-	char buf[buffer_size];
-	const int count = fp->read(buf, sizeof(buf[0]), sizeof(buf) - 1);
+	static char buf[buffer_size];
+	const int count = fp->read(buf, sizeof(buf[0]), sizeof(buf));
```
SheridanR, line 903: if count == sizeof(buf) then buf[count] is an OOB write

SheridanR (approved):
Should fix the OOB write but other than that looks good

src/mod_tools.cpp
WALLOFJUSTICE: fixed

PR #675, opened 2022-09-22
* Text bubbles and new settings for facehotbar

WALLOFJUSTICE:


SheridanR:
seems like this should be gated behind cheats_enabled

WALLOFJUSTICE:
This has skr/bugfixing branch merged in.

Contains new settings and dialogue popups

SheridanR (approved):

PR #783, opened 2023-09-08
Fixes Issues building with steamworks enabled by linking nfd and GTK3.

Iceball457:
For all chunks except the last: The environment variables weren't getting populated correctly so I skipped the command line variables check. This prevents the paths from being set on the command line, but allows them to be correctly read from the environment variables.

For the last chunk: Linked glib2, nfd, and gtk3


SheridanR:
We use this script to create builds in production so I'm very hesitant to merge in anything that can break that.

Iceball457:
As long as it's working for you, there's not really any reason to change it. Frankly I can't get a Windows version to build no matter what I do, cross-compiling or booting into my windows disk

PR #797, opened 2023-10-05
Fix CMakeLists missing GLEW and typos for CMake

philipandresen:
Small fixes required to successfully build on CLion

Most of the diff is the readme I put together to help the next person that tries this. The readme can obviously be excluded if y'all would prefer not to have it on the official repo.

Small fixes implemented based on the advice of my IDE and based on some errors I ran into:
1. Fixed a missing dollar sign on the environment variable reference for GAME_ENABLED
2. Added the `find_package`, `include_directories`, and ` target_link_libraries` calls for GLEW which were previously omitted.
3. Changed INCLUDE(FindZLIB) to find_package(ZLIB QUIET REQUIRED) on the advice of CMake warnings. Functionally it is identical except the `Quiet` and `Required` bits are bonus perks for using the more powerful `find_package`. IMO using "Include" here is like using `eval("x = x + 1")`.

The warning given when using `INCLUDE(FindZLIB)` - possibly due to the [dynamic scoping](https://cmake.org/cmake/help/latest/command/include.html)?
```
CMake Warning (dev) at ...FindPackageHandleStandardArgs.cmake:438 (message):
  The package name passed to `find_package_handle_standard_args` (ZLIB) does
  not match the name of the calling package (PNG).  This can lead to problems
  in calling code that expects `find_package` result variables (e.g.,
  `_FOUND`) to follow a certain pattern.
  ```

### CODE REVIEW

File: cmake/Modules/FindRapidJson.cmake [hunk 1]
```diff
@@ -26,4 +26,4 @@ if (NOT RAPID_JSON_INCLUDE_DIR OR NOT RAPID_JSON_LIBRARIES)
 endif ()
 
 include(FindPackageHandleStandardArgs)
-find_package_handle_standard_args(RAPID_JSON DEFAULT_MSG RAPID_JSON_INCLUDE_DIR )
\ No newline at end of file
+find_package_handle_standard_args(RapidJson DEFAULT_MSG RAPID_JSON_INCLUDE_DIR )
```
philipandresen, line 29: Agh! Forgot to mention this change. It's a good idea to use the package name here spelled and cased the same way it is in the fileName just to make sure there's no confusion. Typically you expect `FindRapidJson` to populate `RapidJson_FOUND` not `RAPID_JSON_FOUND` for example.

Compiler warning:
```
CMake Warning (dev) at D:/Programming/CLion 2023.2.2/bin/cmake/win/x64/share/cmake-3.26/Modules/FindPackageHandleStandardArgs.cmake:438 (message):
  The package name passed to `find_package_handle_standard_args` (RAPID_JSON)
  does not match the name of the calling package (RapidJson).  This can lead
  to problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
  ```

File: .gitignore [hunk 2]
```diff
@@ -75,4 +75,6 @@ steam_appid.txt
 xcode/Barony/Barony.xcodeproj/project.xcworkspace/*
 xcode/Barony/Barony.xcodeproj/xcuserdata/*
 *.ps1
-!LICENSE.nativefiledialog.txt
\ No newline at end of file
+!LICENSE.nativefiledialog.txt
+#Ignore Jetbrains IDE project directories
+/.idea
```
philipandresen, line 80: Hope this isn't a bother! Just helpful to keep IDE stuff out of the project if anyone else uses a JetBrains IDE.

cmake/Modules/FindPNG.cmake
```diff
@@ -9,7 +9,7 @@
 #  PNG_LIBRARY, where to find the PNG library.
 # None of the above will be defined unles zlib can be found.
 # PNG depends on Zlib
-INCLUDE(FindZLIB)
+find_package(ZLIB QUIET REQUIRED)
```
philipandresen, line 12: Caused a compiler warning.
Using "Include" here feels like using `eval("x = x + 1")`. Just a tiny code smell. I didn't find many opinions online but the few I found [tended](https://cgold.readthedocs.io/en/latest/tutorials/cmake-sources/includes.html) to [agree](https://cmake.org/pipermail/cmake/2008-September/024035.html)

The warning given when using `INCLUDE(FindZLIB)`:
```
CMake Warning (dev) at ...FindPackageHandleStandardArgs.cmake:438 (message):
  The package name passed to `find_package_handle_standard_args` (ZLIB) does
  not match the name of the calling package (PNG).  This can lead to problems
  in calling code that expects `find_package` result variables (e.g.,
  `_FOUND`) to follow a certain pattern.
  ```

File: cmake/Modules/FindRapidJson.cmake (hunk 1 above)
Comment (philipandresen) [inline cmake/Modules/FindRapidJson.cmake:29]: Fixes the following Compiler warning:
```
CMake Warning (dev) at .../FindPackageHandleStandardArgs.cmake:438 (message):
  The package name passed to `find_package_handle_standard_args` (RAPID_JSON)
  does not match the name of the calling package (RapidJson).  This can lead
  to problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
  ```
  
The variables automatically set by this call (Eg: RAPID_JSON_FOUND) aren't used - from what I can see - so this change shouldn't have any effect on the build. I tested it with the windows build obviously, but, looking through the build pipeline I didn't see any references that would be broken by this.

damiencarol: will take a look on Linux

PR #817, opened 2023-11-02
Develop (4.1.1)

SheridanR:


WALLOFJUSTICE:
For EOS, have the functions return error codes and set logged in to false when platform handle is null instead of only returning early e.g in EOSFuncs::OnCreateUserCrossplayCallback

PR #823, opened 2024-01-03
Enable compilation without sound

damiencarol:
This simple PR enable compilation without sound (no FMOD or OpenAL required)

*It's part of work to make compilation of source easier in Linux, more PRs will come.*


damiencarol:
> If you can confirm running the game works, then we can merge.

The game run and looks the same, I will lay with it for few weeks but the changes in the code are limited and if you activate FMOD their is not change (only pragmas).

> Also given now that FMOD can be obtained from their website with a free account we've pretty much dropped OpenAL and any code remaining is legacy. (Before we used to use a heavily outdated FMOD version that couldn't be downloaded from FMOD anymore)

Sad to hear that, as I understand FMOD is not free to distribute (as in open-source) so their could be cost induced or even license issues by redistributing.

Also I wanted to work on OpenAL support :/

At least this PR improve the situation and give me first step to enable compilation for some Linux distributions like Arch.

damiencarol:
@WALLOFJUSTICE all good with my tests, I'm able to run and play games without FMOD lib.
is it possible to move forward with this patch? launch CI?

PR #824, opened 2024-01-05
Make BARONY_SUPER_MULTIPLAYER functional

elliot-gh:
This PR fixes issues with the precompiler directive `BARONY_SUPER_MULTIPLAYER` and makes it functional (though still buggy). I do not have Steamworks or EOS so this has only been tested both with multiple local instances and LAN mode with port forwarding over internet.

Fixes:
- Previously, loading an existing save would not work. This is due to `NET_PACKET_SIZE` being larger than the packet data used to load an existing 8 player save. [The `chunk_size` in `lobbyPlayerJoinRequest()` is 98 bytes](https://github.com/TurningWheel/Barony/blob/b5cb1cc5ceb6ef79634bcb17ad86ac9fd46e1e03/src/net.cpp#L1543), and the packet length is defined as `8 + MAXPLAYERS * chunk_size`, resulting in a total of 792 bytes. The old `NET_PACKET_SIZE` is 512, which is too small. I'm not sure what's a good number so I just doubled it.
- There aren't any assets for death boxes past player 5, which resulted in death boxes showing up as doors that would block vision. I simply reused the assets for players 2-4 to make it playable.
- I was having compilation errors with `USE_PLAYFAB` off since I didn't have the libraries/includes, so I moved the includes underneath it.

Bugs:
- There are no assets for death boxes past player 5, so the death box asset colors and player name colors don't match up. I don't have experience with making assets and I don't believe the license granted for this repo allows us to reuse/redistribute assets, so I haven't created new assets.
- Players past player 4 will be invisible on the lobby menu, but there are no issues with character creation or once in game.
- Occasionally there players will desync. I haven't taken a look at how the netcode works to investigate this.

### DIFF
--- a/src/game.hpp
+++ b/src/game.hpp
@@ -35,7 +35,7 @@ class Entity;
 
 #define DEBUG 1
 #define ENTITY_PACKET_LENGTH 46
-#define NET_PACKET_SIZE 512
+#define NET_PACKET_SIZE 1024
 
 // impulses (bound keystrokes, mousestrokes, and joystick/game controller strokes) //TODO: Player-by-player basis.
 extern Uint32 impulses[NUMIMPULSES];
--- a/src/interface/playerinventory.cpp
+++ b/src/interface/playerinventory.cpp
@@ -1985,7 +1985,24 @@ std::string getItemSpritePath(const int player, Item& item)
 			}
 			else
 			{
-				imagePathsNode = list_Node(&items[item.type].images, item.getLootBagPlayer());
+				int playerOwner = item.getLootBagPlayer();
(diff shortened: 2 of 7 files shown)


SheridanR:
Nice work, we have assets for additional player deathboxes (they actually are designed for color remixing) but not XP bars I think. Ghosts need their own assets too.

SheridanR:
We also want to improve the user interface so the lobby browser correctly shows lobbies with greater >4 players, you can adjust the player count properly as a host, kick players, and the default player count remains 4. Raising the player count above that shows a pop-up box talking about instability, game design issues with more than 4 players, etc.

PR #831, opened 2024-03-04
[WIP] Enable back support for OpenAL

damiencarol:
Fix a few misalignments between the code for FMod and OpenAL.

The reason is that the FMod code changed a bit and the part that supports OpenAL was not modified.

**Disclaimer it's a work in progress**
For now, it allows us to compile with OpenAL and have sound support without any crashes (tested on Linux, test for Windows will happen this week).
More commit will be added to this branch to support music.
I'm trying to minimize the changes to have a digest PR, don't shoot me to not fix everything around sound/music :)

What I found:
1) sometimes instead of using `#ifdef SOUND` we need `#ifdef MUSIC`
2) some parts are just missing maintenance, for ex:
   - ex: when some variables were moved for FMod, the variables for OpenAL were not
   - some functions were added for FMod but not for OpenAL (like playSoundNotificationPlayer)

### DIFF
--- a/src/engine/audio/defines.cpp
+++ b/src/engine/audio/defines.cpp
@@ -78,4 +78,6 @@ bool sfxUseDynamicAmbientVolume = true;
 bool sfxUseDynamicEnvironmentVolume = true;
 
 void* fmod_extraDriverData = nullptr;
-#endif //USE_FMOD
+#elif defined OPENAL_ENABLED
+OPENAL_BUFFER** intromusic = nullptr;
+#endif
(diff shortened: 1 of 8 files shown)


damiencarol:
Mmm, the error in the CI is not linked to code compilation error or anything. It seems the CI is broken today :/

> Initiating build-linux_fmod_steam.sh
> Error! No key provided for encrypted dependencies zip. Aborting.
> Error: Process completed with exit code 1.

damiencarol:
@WALLOFJUSTICE it complete now. we can review.

damiencarol:
Added better management when a music fail to load to prevent NPE

Issue #884, opened 2025-05-07
[Linux] Comply with the XDG directory standard


Damglador:
On Linux, the XDG Base Directory Specification[1] defines the paths inside the user's home directory that should be used for different purposes. Barony currently saves the data files in $HOME/.barony which ends up cluttering the home directory even more than it already is or needs to be.

Ideally, cache should be in `$XDG_CACHE_HOME` (`$HOME/.cache`), data in `$XDG_DATA_HOME` (`$HOME/.local/share`) and configs in `$XDG_CONFIG_HOME` (`$HOME/.config`), but everyone dumps everything either in `.config` (Undertale, Unity, Hyper Light Drifter, Vintage Story) or `.loca/share` (Godot, Steam, Mindustry, FTL) anyway. 

[1] https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

PR #886, opened 2025-05-30
Move data directory to ~/.local/share/barony

Damglador:
Fixes TurningWheel/Barony#884

Also adds directories of fmodengine from AUR package to the cmake file.


Damglador:
Ideally this should use XDG_DATA_HOME if possible, perhaps also copy old data from ~/.barony if it exists, and XDG_DATA_HOME/barony doesn't. Any feedback will be appreciated

cmake/Modules/FindFMOD.cmake
shinysocks: what is this?

